PR #165, opened 2015-01-02
Handle 13 digit datetime values

porjo:
Adds support for `datetime` fields that contain 13 digit integers (Unix timestamp with millisecond precision).


mattn:
Thank you

PR #169, opened 2015-01-26
Return error code of sqlite3_step, not sqlite3_reset

mdevan:
Since sqlite3_reset will overwrite the connection-specific last error code, the error code should be collected from sqlite just after the call to sqlite3_step.


mattn:
Thank you

PR #170, opened 2015-01-26
load extensions using the C API instead of SQL

tajtiattila:
This fixes the problem of loading ICU, which needs to replace existing
functions. Replacing existing functions are prohibited when a VM
is running.

I found this problem on Windows x64, debugged by extending icu.c such that it returns a text message.

Another benefit of this change is that the SQL load_extension() function always returns NULL, but
now it is possible to examine the actual return value.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -287,19 +287,13 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 			return nil, errors.New(C.GoString(C.sqlite3_errmsg(db)))
 		}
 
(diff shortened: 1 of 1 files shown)


tajtiattila:
Note that the use of C.sqlite3_load_extension() might not be absolutely necessary: I haven't fully checked, but the SELECT load_extension... might still work without using a prepared statement
(i.e. use conn.Exec directly). Let me know if this should be checked.

I've verified that the calls to C.sqlite3_enable_load_extension() are still necessary. This observation corresponds to the sqlite documentation.

mattn:
Thank you

PR #173, opened 2015-02-11
 optimize create table sql in examples

superalsrk:
Some times in sqlite, the field may be just an integer.

if we use sql like 

```
create table foo (id integer not null primary key, name text, 1 float);
```
it runs wrong, so a better solution is to use sql like

```
create table foo ([id] integer not null primary key, [name] text, [1] float);
```


mattn:
I don't use this syntax without it need to do. Sorry.

PR #183, opened 2015-03-11
Fix gccgo support

stgraber:
Those two commits fix go-sqlite3 building with gccgo by renaming the various files which when compiled using cgo could overwrite sqlite3.c and sqlite3.h and then by passing the extra -lpthread flag to the linker which is required because of the symbols in use.

Travis appears to be happy with this and I've successfully built go-sqlite3 in the following setups:
 - golang-go x86_64 linux
 - golang-go x86 linux
 - golang-go armv7 linux
 - gcc-go arm64 linux
 - gcc-go ppc64le linux


mattn:
Thanks!

PR #194, opened 2015-04-10
Add a txlock option when opening databases

### LINKED ISSUE #189: Support 'BEGIN IMMEDIATE' ?
Hi,

We'd like to use 'BEGIN IMMEDIATE' for our transactions.  I don't see a clean way to support this through database/sql, certainly at a per-transaction basis.  For us it would be ok to just have 'Open(file, "path?_begin_immediate")' mean all db.Begin()s would do "BEGIN IMMEDIATE".  But if there is a cleaner way I'm open to that.

If the Open() way is ok with you I can send a pull request.

hallyn:
When specified, changes the default locking at a tx.Begin.

Closes #189

Signed-off-by: Serge Hallyn <serge.hallyn@ubuntu.com>

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -117,8 +117,9 @@ type SQLiteDriver struct {
 
 // Conn struct.
 type SQLiteConn struct {
-	db  *C.sqlite3
-	loc *time.Location
+	db     *C.sqlite3
+	loc    *time.Location
+	txlock string
 }
 
 // Tx struct.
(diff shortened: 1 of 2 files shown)


mattn:
Could you please add test?

hallyn:
I'm not sure how I would test for it.  Or do you mean just make
sure that it doesn't crash with various ?_txlock= options?

mattn:
Yes, simple passing true case, I want.

On 4/13/15, Serge Hallyn <notifications@github.com> wrote:
> I'm not sure how I would test for it.  Or do you mean just make
> sure that it doesn't crash with various ?_txlock= options?
>
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/mattn/go-sqlite3/pull/194#issuecomment-92337617


-- 
- Yasuhiro Matsumoto

mattn:
Thank you

PR #196, opened 2015-04-12
Fix NULs in text.

### LINKED ISSUE #195: Convert BLOB to image
I'm trying to create an mbtiles server in Go. The file follows the mbtiles spec. The `tile_data` field is a BLOB and i'm trying to query the database and get the corresponding images (stored as blob types).
I'm using this sqlite3 driver to query the database and i'm not getting the expected result.

This [Stackoverflow thread](http://stackoverflow.com/q/29452538/1373105) explains better my problem with some example code and image for clarification.

I even ask this at the [goolang-nuts google group](https://groups.google.com/d/topic/golang-nuts/DEkLPMx77-c/discussion) and someone suggest that this could be an issue with the sqlite driver.

Any help appreciated.

egonelbre:
NUL character is a valid symbol in UTF8.

Fixes #195

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -624,7 +624,9 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 		case C.SQLITE_TEXT:
 			var err error
 			var timeVal time.Time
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #207, opened 2015-06-05
Several fixes, mostly to tests

larsmans:
Here's a series of small patches. The main one makes the tempfile generation in the tests more robust, as well as making the files easier to spot afterwards (the tests tend to leave a lot of tempfiles lying around). This might just solve the impossible-to-reproduce test failures that I've been seeing.


larsmans:
Hm, Travis seems to have stumbled on one of those failures in https://travis-ci.org/mattn/go-sqlite3/builds/65578505...

mattn:
merged. thanks.

Issue #148, opened 2014-09-24
Help for sql error "database is locked"


pantaluna:
Hi,

```
sqlite3.Error{Code:5, ExtendedCode:5, err:"database is locked"}
```

I have written an HTTP API server in Go that uses SQLite3 and your great package to implement data persistence.

The daemon works fine on the dev machine but once it is deployed on a test server and put under a slightly heavier load (100 API requests / second via HTTP) then I noticed that the API server throws SQL error <database is locked>. The sql errors occur randomly and in less then 0.2% of the transactional statements.

Some stats of a stress test:

```
Date            #Completed      #Errors     Ratio%
21:00           384215          628         0.163%
22:00           525816          744         0.141%
23:00           687652          812         0.118%
```

I have narrowed the problem down to the SQLite3 database (and maybe how I use it) when doing Insert, Update and Delete (Select works fine if no program is doing transactions and is blazingly fast).

The sql error <database is locked> starts popping up as soon as the execution of an sql statement has to wait more than 5 seconds. 

=> So I have made a small test program to make it easier for you to reproduce it.
https://gist.github.com/pantaluna/89ffe94ec30fb28d9fb8

```
# Script
go get
go run databaselocked.go --addtestdata
go run databaselocked.go --updates
go run databaselocked.go --updates > updates.log 2>&1
grep "ERROR" updates.log
```

I noticed that the 5 second timeout is hardcoded in the go-sqlite3 package. I do not see increasing the timeout setting in the go-sqlite3 package as a real solution, nor throttling down the number of active sql connections to 1 (that is possible in database/sql).

I think the whole database is locked when an sql client executes a (simple) transactional statement; even if the Updates occur on different tables, or on the same table but different rows (the latter is implemented in the test program).

Can you please advice me on how to use the go-sqlite3 package correctly, and avoid those random <database is locked> errors?

Thank you.

wuranbo:
Looks it is not fixed. 
I got he exactly same error.
When I set "cache=shared&mode=rwc", same error described upon...

origsun:
Old thread but this might help someone:

Running the gist in @pantaluna 's comment ( https://gist.github.com/pantaluna/7ad0988269119507539b ) still produces the same error (`database table is locked: counters`) on my machine, but setting `gDbMaxOpenConns` to `1` works fine: no errors and no obvious slowing down.

A related issued: https://github.com/applikatoni/applikatoni/issues/35

mattn:
[...]

PR #211, opened 2015-06-17
*fix) add define HAVE_USLEEP

wuranbo:
It's what I used for the issue #148 
Based on this blog http://beets.radbox.org/blog/sqlite-nightmare.html, http://mail-index.netbsd.org/current-users/2012/06/01/msg020320.html.
It works fine to me now, but I am very NOT sure it is the code what solved the problem.

Just PR for more discuss.The #148 has obstruct me days.

### DIFF
--- a/sqlite3-binding.h
+++ b/sqlite3-binding.h
@@ -7474,9 +7474,8 @@ struct sqlite3_rtree_query_info {
 }  /* end of the 'extern "C"' block */
 #endif
 
+#endif  /* ifndef _SQLITE3RTREE_H_ */
+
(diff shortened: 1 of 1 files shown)


donlzx:
This works for my issue (database is locked). Thanks!

mattn:
do NOT define in sqlite3-binding.c 
please add cgo CFLAGS.

wuranbo:
@mattn got it. @donlzx attention buddy.

ligustah:
+1

I ran into this error as well.

Issue #213, opened 2015-06-25
Is there any way to use the driver with SQLITE_ENABLE_UPDATE_DELETE_LIMIT flag?


elgris:
This flag is useful when one is executing DELETE query with LIMIT clause: `DELETE FROM "sometable"  WHERE (name = 'foo') LIMIT 5`.
By default the driver returns an error: `near "LIMIT": syntax error`.

Is there any way to enable it for the driver? There is one possible problem: documentation says that this flag cannot be used with "amalgamation" source of sqlite3 driver: https://www.sqlite.org/compile.html

gregf:
:+1:

mattn:
anyone have objection?

PR #215, opened 2015-06-26
Updated SQLite bindings to support SQLITE_ENABLE_UPDATE_DELETE_LIMIT flag

elgris:
Supports #213.

Please consider this PR as a major change because it updates SQLite bindings to version 3.8.10.2.
The amalgamation source code was built with SQLITE_ENABLE_UPDATE_DELETE_LIMIT flag enabled.

### DIFF
--- a/sqlite3-binding.h
+++ b/sqlite3-binding.h
@@ -43,21 +43,25 @@ extern "C" {
 
 
 /*
-** Add the ability to override 'extern'
+** Provide the ability to override linkage features of the interface.
 */
 #ifndef SQLITE_EXTERN
 # define SQLITE_EXTERN extern
 #endif
-
 #ifndef SQLITE_API
 # define SQLITE_API
 #endif
-
+#ifndef SQLITE_CDECL
+# define SQLITE_CDECL
+#endif
+#ifndef SQLITE_STDCALL
+# define SQLITE_STDCALL
+#endif
 
 /*
 ** These no-op macros are used in front of interfaces to mark those
 ** interfaces as either deprecated or experimental.  New applications
-** should not use deprecated interfaces - they are support for backwards
+** should not use deprecated interfaces - they are supported for backwards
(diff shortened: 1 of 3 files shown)


mattn:
Please don't upgrade sqlite3.c. go has problem about this.

https://github.com/golang/go/issues/9356

elgris:
Hm... Is it because of version 3.8.10.2 or there is another reason and current `squirrel-binding.c` is highly customized?Is it ok to rebuild `sqlite.c` v3.8.5 (as it is in master right now), but with SQLITE_ENABLE_UPDATE_DELETE_LIMIT enabled?

mattn:
Yes, please.

elgris:
Ok, I rolled back bindings to v3.8.5. Could you please check failing scenario on Windows?

mattn:
Hmm, totally diff seems deleted/appended into sqlite3-binding.c. I'm thinking we can't touch sqlite3-binding.c for now.

elgris:
Actually, this is rebuild version of `sqlite3-binding.c`. I build it from the sources of version 3.8.5 (like it was before), just added SQLITE_ENABLE_UPDATE_DELETE_LIMIT flag at configure and build steps. Generally, this is the same code, just with extra function added

gregf:
@mattn any word on this? I could really use it.

mattn:
https://www.sqlite.org/compile.html

> If this option is defined, then it must also be defined when using the 'lemon' tool to generate a parse.c file. Because of this, this option may only be used when the library is built from source, not from the amalgamation or from the collection of pre-packaged C files provided for non-Unix like platforms on the website.

Is this your expected?

gregf:
@mattn yes

mattn:
Could you please rebase master?

GJRTimmer:
Should be closed because of #564

mattn:
note that not mean that this enabling delete limit also when using libsqlite3 on system always.

PR #225, opened 2015-08-07
Fix for go-sqlite3 truncating 64-bit lastInsertIDs on 32-bit systems

kiwih:
Currently, when using go-sqlite3 on 32 bit systems, LastInsertId is truncated from 64 bits to 32 bits, before being cast back to 64 bits.

This happens because golang's C.long is implementation dependent, and on 32-bit systems is 32 bits.

I have changed this from C.long to C.longlong.

Go test results:
```
BenchmarkExec           500000     342197 req/s
BenchmarkQuery          200000      88448 req/s
BenchmarkParams         100000      72247 req/s
BenchmarkStmt           200000     106827 req/s
BenchmarkRows             3000       2087 req/s
BenchmarkStmtRows         3000       2144 req/s
PASS
ok      github.com/kiwih/go-sqlite3     31.694s
```

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -48,21 +48,21 @@ _sqlite3_bind_blob(sqlite3_stmt *stmt, int n, void *p, int np) {
 #include <stdint.h>
 
 static int
-_sqlite3_exec(sqlite3* db, const char* pcmd, long* rowid, long* changes)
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

PR #229, opened 2015-08-21
Implement support for calling Go functions from SQLite

### LINKED ISSUE #226: Support for UDFs?
For me, one of the most interesting features of SQLite, is the ability to extend the behaviour of SQL queries by implementing simple or aggregate UDFs (user-defined functions).

They are critical to implement the REGEXP operator for instance, and, in my current scenario, a RANK() function that operates on data returned by the FTS matchinfo().

I don't think I've seen anything regarding this in the docs (in pure Go). Would it be even possible to do it without C?

danderson:
Fixes #226.

Supports scalar functions and aggregation functions.

I've benchmarked the call stack, and even with the reflection, the performance matches the QueryRow benchmark (18 microseconds per "SELECT custom_func(1,2)").

### DIFF
--- a/_example/go_custom_funcs/main.go
+++ b/_example/go_custom_funcs/main.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	"log"
+	"math"
+	"math/rand"
+
+	sqlite "github.com/mattn/go-sqlite3"
+)
+
+// Computes x^y
+func pow(x, y int64) int64 {
+	return int64(math.Pow(float64(x), float64(y)))
+}
+
+// Computes the bitwise exclusive-or of all its arguments
+func xor(xs ...int64) int64 {
+	var ret int64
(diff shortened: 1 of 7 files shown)


mattn:
Oh, awesome. I'll look into it in later.

danderson:
I also have ideas for improvements to this code (do more work in register, less work in the actual call, and support more data types), but the change was becoming quite big, so if you're okay with it, I'll send the improvements in separate changes/pull requests.

danderson:
Ping?

mattn:
I'll merge this after CI. :+1:

mattn:
Could you handle error on CI?

danderson:
Sorry, there was a typo in one of the tests. Fixed and repushed, CI passes now.

mattn:
Coooooooooool! :+1:

Dashed:
This is slick. :+1:

PR #235, opened 2015-09-04
remove dlopen if static_build flag

jfrazelle:
This patch adds a CFLAG `-DSQLITE_OMIT_LOAD_EXTENSION` to omit dlopen 
if go build flag `static_build` is passed.

we use this flag on docker/docker, but we can change the name or something 
if you prefer something else.

we do not use the extensions and this will remove the dlopen warning 
when we compile static binaries

cc @tianon

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -355,23 +355,8 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 	conn := &SQLiteConn{db: db, loc: loc, txlock: txlock}
 
 	if len(d.Extensions) > 0 {
-		rv = C.sqlite3_enable_load_extension(db, 1)
-		if rv != C.SQLITE_OK {
-			return nil, errors.New(C.GoString(C.sqlite3_errmsg(db)))
-		}
-
-		for _, extension := range d.Extensions {
-			cext := C.CString(extension)
-			defer C.free(unsafe.Pointer(cext))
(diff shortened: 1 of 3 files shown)


jfrazelle:
I dont really like how i named the files tbh so open to opinions

kolyshkin:
Name ``static_build`` is sort of too broad. Even in case of static build, one might still want to use ``dlopen()`` -- and the linker is not against it, it just lets you know (with a warning) that the binary is not truly static.

So, I'd rename the build tag to ``sqlite_omit_load_extension``, as thay way it will be 1:1 mapping to the corresponding C flag ``SQLITE_OMIT_LOAD_EXTENSION``, which is already documented. So, if you don't want sqlite's LOAD extension, you specify this C flag (or this Go build tag), and it's irrelevant why you don't want this extension -- to get rid of a linker warning, to make your binary a bit smaller, or perhaps just because you are paranoid about security.

jfrazelle:
sgtm!

kolyshkin:
Otherwise LGTM

jfrazelle:
updated!

mattn:
Thanks.

tianon:
@mattn would you mind making a new release tag including this commit? :innocent: :heart:

mattn:
done https://github.com/mattn/go-sqlite3/tree/v1.1.0

PR #241, opened 2015-09-20
Add FTS4 unicode61 tokenizer support

looi:
This adds support for the FTS4 unicode61 tokenizer, which requires no external dependencies and is enabled by default in the binaries on sqlite.org. I also added a test to ensure that the unicode61 tokenizer is working - a query for "fevrier" should match documents containing the word "février"


mattn:
merged b808f01f660667b1510acf216aa2b9fe7f555ddb

PR #243, opened 2015-10-10
Store/retrieve timezones for time.Time values.

augustoroman:
Previously, the timezone information for a provided value was discarded
and the value always stored as in UTC.  However, sqlite allows specifying
the timezone offsets and handles those values appropriately.  This change
stores the timezone information and parses it out if present, otherwise
it defaults to UTC as before.

One additional bugfix:  Previously, a unix timestamp in seconds was
parsed in the local timezone (rather than UTC), in contrast to a unix
timestamp in milliseconds that was parsed in UTC.

While fixing that extra bug, I cleaned up the parsing code -- no need to
convert to a string and then parse it back again and risk a parse error,
just to check the number of digits.

The tests were extended to cover non-UTC timezones storage & retrieval,
meaningful unix timestamps, and correct handling of a trailing Z.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -99,6 +99,10 @@ import (
 // into the database. When parsing a string from a timestamp or
 // datetime column, the formats are tried in order.
 var SQLiteTimestampFormats = []string{
+	// By default, store timestamps with whatever timezone they come with.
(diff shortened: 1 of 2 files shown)


mattn:
Thanks

PR #250, opened 2015-10-31
Update README. Install go-sqlite3 to avoid builds with gcc

DavidGamba:
I was struggling with the build speeds of using gcc. My editor will compile the code on every save to check for syntax errors. By installing, my app builds fast again.


mattn:
Thanks.

PR #252, opened 2015-11-06
Fix README project address

marcelolima:
Missing a 3 in the https://github.com/mattn/go-sqlite3 project address.


mattn:
thank you

PR #253, opened 2015-11-09
Implement read-only mode via _readonly=true flag in Open()

Tasssadar:


mattn:
Thank you. Could you please add test?

Tasssadar:
Oh, never mind, this can already be done via Open("file:something.sqlite?mode=ro"), it is even in the tests: https://github.com/mattn/go-sqlite3/blob/21637a6531000298cf4a2d622e7d049a289df9ef/sqlite3_test.go#L89

mattn:
Oh, sorry.

Issue #222, opened 2015-07-30
Ship with SQLite 3.8.11 Bindings


alixaxel:
SQLite 3.8.11 has just been released, and according to the [changelog](https://www.sqlite.org/releaselog/3_8_11_1.html) it comes with some nice performance improvements.

It would be great if the package was updated to use this version.

mattn:
https://github.com/mattn/go-sqlite3/pull/171

This is same issue related on https://github.com/golang/go/issues/9356

I'll wait release of go1.5

ayanamist:
go 1.5 has been released for a while, so?

mattn:
will do but please wait.

rusco:
I propose setting up a Gratipay account for this !

alixaxel:
@rusco Good idea, :+1:.

mattn:
I didn't merge this PR, but I could decide to merge latest amalgamation code to latest. thanks!
#264

alixaxel:
@mattn Amazing, thanks!

rusco:
:clap:

PR #171, opened 2015-02-10
SQLite amalgamation to 3.8.9

denji:
The new release also conducted performance optimization of full-text queries fts3/fts4, used in combination with the **OR** operator. Fixed a potential integer overflow in functions `sqlite3_blob_read()` and `sqlite3_blob_write()`.

Before (`go test -bench=. -timeout=30s`):
```
BenchmarkExec           200000     107113 req/s
BenchmarkQuery           50000      34443 req/s
BenchmarkParams          50000      29273 req/s
BenchmarkStmt           100000      49890 req/s
BenchmarkRows             2000       1059 req/s
BenchmarkStmtRows         2000       1088 req/s
PASS
ok  	_/tmp/go/go-sqlite3-master	16.157s
```

After:
```
BenchmarkExec           200000     110548 req/s
BenchmarkQuery           50000      35105 req/s
BenchmarkParams          50000      29825 req/s
BenchmarkStmt           100000      51111 req/s
BenchmarkRows             2000       1076 req/s
BenchmarkStmtRows         2000       1100 req/s
PASS
ok  	_/tmp/go/go-sqlite3-sqlite388	15.905s
```

PR #264, opened 2016-01-22
Upgrade amalgamation 3.10.2

vovkasm:
Upgrade embedded sqlite sources to version 3.10.2
This actualize #171 and #222 
Tested with go version go1.5.3 darwin/amd64


mattn:
Now, I decided to upgrade amalgamation code to latest because most of users doesn't use go1.4 at least.

mattn:
thank you

PR #267, opened 2016-01-29
bind: pass &v[0] in direct call to C

ianlancetaylor:
In Go 1.6, the cgo checking rules are more precise when they see an
address operation as an argument to the C function.  When you pass &v[0]
to a C function, the cgo check just verifies that v itself does not
contain any pointers.  When you write `p := &v[0]` and then pass p to
the C function, the cgo check is conservative: it verifies that the
entire memory block to which p points does not contain any pointers.
When the bind function is called by code that passes a slice that is
part of a larger struct, this means that the cgo check will look at the
entire larger struct, not just the slice.  This can cause a surprising
run time failure.

Avoid this problem by rewriting the code slightly to pass &v[0] in the
call to the C function itself.

In particular this fixes the tests of github.com/jmoiron/sqlx when using
Go 1.6.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -814,11 +814,11 @@ func (s *SQLiteStmt) bind(args []driver.Value) error {
 		case float64:
 			rv = C.sqlite3_bind_double(s.s, n, C.double(v))
 		case []byte:
-			var p *byte
-			if len(v) > 0 {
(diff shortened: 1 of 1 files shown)


mattn:
I see. Thank you.

PR #268, opened 2016-01-29
callback: use handles rather than passing Go pointers

ianlancetaylor:
The cgo pointer passing rules forbid passing a Go pointer to C if that
pointer points to memory containing other Go pointers.  This is true
even if the Go pointer is converted to uintptr.

This change fixes the code to use a handle instead, and to look up the
handle in the callback function.

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -24,27 +24,73 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"sync"
 	"unsafe"
 )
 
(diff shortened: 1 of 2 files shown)


mattn:
Somehow, I had expected the patch. And also I just sure that the fixes against cgoCheckPointer will be large for another projects too. Thank you very much. And Sorry for taking your precious time.

PR #278, opened 2016-02-28
Conditional build for the JSON1 Extension

rkintzi:
Use ``-tags json1`` to enable [JSON1 Extension](http://www.sqlite.org/json1.html)


mattn:
Thanks.

PR #279, opened 2016-02-28
Source and destination are reversed

otoolep:
While examining this backup code, it seems to me that the use of `srcDb` and `destDb` is backwards since `destDb` was the source of the backup and `srcDb` was the destination of the backup.

I have reversed these names.


otoolep:
Perhaps I am missing something, but I think this is what is actually meant by this example code.

mattn:
Sorry delay. Thanks.

PR #291, opened 2016-03-24
Fix sqlite "regexp" function name in documentation

diocles:
Just a drive-by typo fix that bit me when I tried to use this example to add support for the REGEXP operator!


mattn:
Thank you

PR #294, opened 2016-04-02
Move sqlite3 amalgation files a directory up.

### LINKED ISSUE #293: code/ subdirectory is not properly handled by the go tool
The sqlite3 amalgamation files are ignored by the go tool. It does not cause any issue during the buils as these files are included by higher level other C files, but it prevent to easily vendor this package with the help of godep for instance. The command go list -json should list these file somehow.

dgsb:
The purpose is to ease the use of vendoring files like godep.
The C sqlite3 files have been added a go compilation conditional flag
Fix #293

### DIFF
--- a/backup.go
+++ b/backup.go
@@ -6,7 +6,11 @@
 package sqlite3
 
 /*
+#ifndef USE_LIBSQLITE3
 #include <sqlite3-binding.h>
+#else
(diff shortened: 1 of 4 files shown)


mattn:
Thanks.

PR #296, opened 2016-04-14
add missing error checks in simple example

stanim:
Any error encountered during iteration should be handled properly, see:
https://golang.org/pkg/database/sql/#Rows
http://go-database-sql.org/retrieving.html


mattn:
thank you

stanim:
You're welcome.

Issue #261, opened 2015-12-27
Add support for overriding the entrypoint for sqlite3_load_extension()


weitzj:
The current plugin supports loading extensions  as seen in: https://github.com/mattn/go-sqlite3/blob/master/sqlite3_load_extension.go#L28 using:

```
C.sqlite3_load_extension(c.db, cext, nil, nil)
```

whereas the method is defined as (https://www.sqlite.org/c3ref/load_extension.html)

```
int sqlite3_load_extension(
  sqlite3 *db,          /* Load the extension into this database connection */
  const char *zFile,    /* Name of the shared library containing extension */
  const char *zProc,    /* Entry point.  Derived from zFile if 0 */
  char **pzErrMsg       /* Put error message here if not 0 */
);
```

I would like to have/provide support for overriding the third argument, the entrypoint to the extension.

My problem:

I want to load the `spatialite` extension via `go-sqlite3` on Android. `spatialite` comes in 2 flavours, either as a shared library (`libspatialite.so`) or as an sqlite3 extension `mod_spatialite.so`. Since `mod_spatialite.so` has to be built with Android `ndk-build`, and `ndk-build` **only supports library names of the form `libMODULENAME.so`**, the resulting spatialite extension filename is called `libspatialite.so`, whilst the entrypoint still points to the original modulename, `mod_spatialite.so` and thus will be `sqlite3_modspatialite_init`.

Sqlite will automatically guess an entrypoint given a filename, if no entrypoint is given:

- `libspatialite.so` has the (guessed) entrypoint `sqlite3_spatialite_init`
- `mod_spatialite.so` has the (guessed) entrypoint `sqlite3_modspatialite_init`

What I need:

`load_extension('libspatialite.so','sqlite3_modspatialite_init')`

I do not know, what would be a good variant how to integrate this in `go-sqlite3`. Maybe an `Extension` struct.

mattn:
How about this? go-sqlite3 make public name of loadExtensions to LoadExtensions. And add one more argument to specify proc name. Then users can use it like below.

```
conn.LoadLibrary("foo.so", "bar_init")
```

mattn:
closed by #298

shaxbee:
@weitzj: FYI - I've added android spatialite extension loading to [shaxbee/go-spatialite](https://github.com/shaxbee/go-spatialite)

weitzj:
@shaxbee  awesome.thanks

mattn:
[...]

PR #298, opened 2016-04-18
Expose LoadExtension with entry point

shaxbee:
Expose sqlite3_load_extension with entry point argument to allow loading extensions where library name does not match entry point name, eg. libspatialite.

Solves #261

### DIFF
--- a/sqlite3_load_extension.go
+++ b/sqlite3_load_extension.go
@@ -37,3 +37,27 @@ func (c *SQLiteConn) loadExtensions(extensions []string) error {
 	}
 	return nil
 }
+
+func (c *SQLiteConn) LoadExtension(lib string, entry string) error {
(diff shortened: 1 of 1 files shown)


mattn:
thank you

shaxbee:
Please put it on hold - I'm testing it on darwin right now and it seems that system installed libsqlite3 *does not* support extension loading.
Shall we require darwin users to use homebrew version of it which includes extension loading? If so I have a change ready for that.

shaxbee:
It seems that forcing brew library is an only option as otherwise the go-sqlite3 will not build on darwin complaining on extension functions.

weitzj:
@shaxbee See https://github.com/weitzj/libspatialite to get a spatialite library for Android.

shaxbee:
@weitzj Thanks for the link, currently I'm not building anything for Android, just wanted to solve #261 :-)

PR #300, opened 2016-04-18
Fix inconsistent tx state with database/sql.

### LINKED ISSUE #184: inconsistent transaction state
First of all，I open the db file in the following way.
```go
db, err := sql.Open("sqlite3", "file:locked.sqlite?cache=shared&mode=rwc")
```

My program commit a db transaction **periodically**. At the same time, another program(like /usr/bin/sqlite3) do a big query on the same db. Then error reported is:
```bash
2015/03/16 20:05:49 "database is locked": Commit
# Error from the successive transation
2015/03/16 20:05:49 "cannot start a transaction within a transaction": Begin transaction
```
IMO, a fail-commited transaction means a rollbacked transaction. Am I wrong ?
I do the rollback manually. And error reported became:
```bash
2015/03/16 20:06:51 "database is locked": Commit
2015/03/16 20:06:51 "sql: Transaction has already been committed or rolled back": Rollback
2015/03/16 20:06:51 "cannot start a transaction within a transaction": Begin transaction
```
Now, I‘m confusing. What should I do to begin a new transaction? The old transaction is rolled back, **OR NOT**?

Do not inform me: **Close-Then-ReOpen the db**. It's just a workaround.

> ## func Open
```go
func Open(driverName, dataSourceName string) (*DB, error)
```
> Open opens a database specified by its database driver name and a driver-specific data source name, usually consisting of at least a database name and connection information.

> Most users will open a database via a driver-specific connection helper function that returns a *DB. No database drivers are included in the Go standard library. See http://golang.org/s/sqldrivers for a list of third-party drivers.

> Open may just validate its arguments without creating a connection to the database. To verify that the data source name is valid, call Ping.

> **The returned DB is safe for concurrent use by multiple goroutines and maintains its own pool of idle connections. Thus, the Open function should be called just once. It is rarely necessary to close a DB.**

Finally I need help and also want to help.

sqweek:
The semantics of sql.Tx.Commit impose that the transaction is
finished and cleaned up by the time the driver's Commit function
returns. However sqlite3 leaves the transaction open if COMMIT
fails due to an SQLITE_BUSY error, so *we* must clean it up.

Closes #184.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -292,6 +292,12 @@ func (ai *aggInfo) Done(ctx *C.sqlite3_context) {
 // Commit transaction.
 func (tx *SQLiteTx) Commit() error {
 	_, err := tx.c.exec("COMMIT")
+	if err != nil && err.(Error).Code == C.SQLITE_BUSY {
+		// sqlite3 will leave the transaction open in this scenario.
+		// However, database/sql considers the transaction complete once we
+		// return from Commit() - we must clean up to honour its semantics.
+		tx.c.exec("ROLLBACK")
+	}
 	return err
 }
 


sqweek:
To test the change I used this little program: https://play.golang.org/p/wnMn2OVhfn

1. Run the program a few times to create the database (`/tmp/issue184.db`) and insert some rows
2. In a separate terminal run the commands:
 1. `sqlite3 /tmp/issue184.db`
 2. `sqlite> BEGIN;`
 3. `sqlite> SELECT * FROM tab;`
3. The sqlite3 process has a read-lock on the database. Run the program again.
4. The program will get a `database is locked` error trying to COMMIT, as expected.
5. The program sleeps for 5 seconds and tries to `tx.Commit()` again, which fails because `database/sql` considers the transaction finished
6. The program tries to `tx.Rollback()` which fails for the same reason
7. The program calls `db.Begin()` to open a new transaction, which succeeds

Before the change, the `db.Begin()` call failed with error `cannot start a transaction within a transaction`.

I'm not sure how easy it would be to isolate this into a test case.

mattn:
Thanks.

PR #301, opened 2016-04-22
Ubuntu 14.04 LTS libsqlite3 compatibility

shaxbee:
When testing go-sqlite3 with libsqlite3 on Travis-CI in Trusty environment (Ubuntu 14.04.2 LTS) I found out that symbol SQLITE_DETERMINISTIC is not defined in sqlite 3.8.2.

To solve that issue i default SQLITE_DETERMINISTIC to 0 if it is not defined.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 language: go
 go:
+  - 1.5
+  - 1.6
   - tip
 before_install:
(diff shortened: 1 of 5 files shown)


shaxbee:
Need to rework travis build for OS X. Closing for now..

PR #322, opened 2016-08-20
Add some testing of the backup functionality.

a-p-:
This adds some testing of the backup functionality.

It tests using a single `Step(-1)` call, as well as multiple `Step()` calls.

It also does some testing of the `PageCount()` and `Remaining()` methods.


mattn:
Cool.

PR #323, opened 2016-09-03
Make Go 1.7 fail compilation without cgo

FiloSottile:
The package is useless without cgo, but Go 1.7 will happily build it
since doc.go is a pure-Go file, even if CGO_ENABLED=0, like when
cross-compiling.

This is particularly problematic since the package is usually imported
for its side effects, and users would successfully build a broken binary.

Workaround golang/go#16981

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -110,3 +110,5 @@ See the documentation of RegisterFunc for more details.
 
 */
 package sqlite3
+
+import "C"


FiloSottile:
Thinking about it, we can make it even more useful by showing a custom error from a file with a nocgo build tag.

mattn:
Thanks. Let's doing another issue. :+1:

Issue #332, opened 2016-09-28
cannot build  go-sqlite3 in macOS 10.12


Macbets:
i tried use go install github/mattn/go-sqlite3 and get bad strings in build log 

../../.go/src/github.com/mattn/go-sqlite3/sqlite3-binding.c:20515:17: warning: 'OSAtomicCompareAndSwapPtrBarrier' is deprecated: first deprecated in macOS 10.12 - Use atomic_compare_exchange_strong() from <stdatomic.h> instead [-Wdeprecated-declarations]
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/libkern/OSAtomicDeprecated.h:547:6: note: 'OSAtomicCompareAndSwapPtrBarrier' has been explicitly marked deprecated here

mattn:
This is not an issue of go-sqlite3. Please report this to the author of sqlite3

nyarly:
I've just reported this to the sqlite3 authors, who's position is roughly "wontfix" on the grounds that a) it's a harmless warning and that b) Apple makes use of the OSAtomicCompareAndSwapPtrBarrier function in their own code, so the deprecation is unlikely to become a removal any time soon and c) compiling with -DSQLITE_WITHOUT_ZONEMALLOC (which is there for 10.2 support) removes the call regardless if there's a concern about being caught out without the function call.

So, upshot is that SQLite isn't going to make a change here for the forseeable future.

That said, if there were a way to suppress the warning during builds? If nothing else, it pollutes every `go test` that touches go-sqlite3. I don't know enough about how cgo works to recommend a concrete solution.

mattn:
If put `-DDSQLITE_WITHOUT_ZONEMALLOC` into sqlite3.go, do this fix your issue? Sorry I don't have mac.

nyarly:
It might fix the issue *if* at some future point the deprecated function were deleted by Apple, which sounds unlikely.

I think the PR (merged already) has a better solution for the time being. And the foreseeable future.

jadeydi:
I fixed this problem, by following steps:

1. update homebrew
2. update sqlite
3. git pull github.com/mattn/go-sqlite3

U can try this.

puzanov:
Had the same issue. Here is how I fixed it.
$ brew update
$ brew reinstall sqlite
$ go get -v -u github.com/mattn/go-sqlite3

Issue #328, opened 2016-09-15
golang 1.7.1 deprecation warning


stephenhu:
running on mac os x, recently downloaded go 1.7.1, on build i get the following warning which doesn't block progress, but certainly should be something that needs to be addressed.

this warning doesn't appear with go 1.6.3

mbahu:gostats hu$ go get github.com/mattn/go-sqlite3
# github.com/mattn/go-sqlite3
In file included from ../../mattn/go-sqlite3/sqlite3-binding.c:2:
../../mattn/go-sqlite3/code/sqlite3-binding.c:18400:17: warning: 'OSAtomicCompareAndSwapPtrBarrier' is deprecated: first deprecated in macOS 10.12 - Use atomic_compare_exchange_strong() from <stdatomic.h> instead [-Wdeprecated-declarations]
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/libkern/OSAtomicDeprecated.h:547:6: note: 'OSAtomicCompareAndSwapPtrBarrier' has been explicitly marked deprecated here

mattn:
send report this issue to author of sqlite3. this is not an issue of go-sqlite3.

nyarly:
Repeated in #332

Issue #311, opened 2016-06-17
warning: 'OSAtomicCompareAndSwapPtrBarrier' is deprecated


monoflash:
```
github.com/mattn/go-sqlite3/sqlite3-binding.c:18709:17: warning: 'OSAtomicCompareAndSwapPtrBarrier' is deprecated: first deprecated in OS X 10.12 - Use atomic_compare_exchange_strong() from <stdatomic.h> instead [-Wdeprecated-declarations]
/usr/include/libkern/OSAtomicDeprecated.h:547:6: note: 'OSAtomicCompareAndSwapPtrBarrier' has been explicitly marked deprecated here
```

This warning displayed in macOS Sierra (v. 10.12)

mattn:
please your patch to the author of sqlite3.

objectivesheep:
Any way to bypass sqlite3?

nyarly:
Repeated in #332

hktalent:
mac os 10.12.3
cd /usr/local/opt/go/src/github.com/mattn
git clone https://github.com/mattn/go-sqlite3.git
cd go-sqlite3
git pull

cd /usr/local/opt/go/src/github.com/future-architect/vuls
make install
...........
[INFO]	--> Exporting golang.org/x/crypto
[INFO]	--> Exporting github.com/parnurzeal/gorequest
[INFO]	--> Exporting golang.org/x/sys
[INFO]	--> Exporting github.com/rifflock/lfshook
[INFO]	--> Exporting github.com/moul/http2curl
[INFO]	Replacing existing vendor dependencies
go install -ldflags "-X 'main.version=v0.2.0' -X 'main.revision=1e8f24d'"
# github.com/future-architect/vuls/vendor/github.com/mattn/go-sqlite3
In file included from vendor/github.com/mattn/go-sqlite3/sqlite3-binding.c:2:
vendor/github.com/mattn/go-sqlite3/code/sqlite3-binding.c:16475:17: warning: 'OSAtomicCompareAndSwapPtrBarrier' is deprecated: first deprecated in macOS 10.12 - Use atomic_compare_exchange_strong() from <stdatomic.h> instead [-Wdeprecated-declarations]
/usr/include/libkern/OSAtomicDeprecated.h:547:6: note: 'OSAtomicCompareAndSwapPtrBarrier' has been explicitly marked deprecated here

domino14:
this is making Go compilation extremely slow. anyone know what to do?

[...]

PR #333, opened 2016-10-04
Added CFLAG config to ignore deprecations

nyarly:
Added a CFLAG line to sqlite3.go that prevents the compiler from emitting warnings about deprecated function calls - addresses #332 #328 #311


mattn:
Ah, I notify this just now. Will look.

mattn:
Thanks.

PR #345, opened 2016-10-27
fix double free in SQLiteBackup.Close() on error

joeshaw:
The sqlite3_backup_finish() function never fails, it just returns the
error code from previous operations.  Previously if it returned an
error, the finalizer wouldn't be unset and sqlite3_backup_finish() would
be run again on an already-finished backup.  This results in a
double-free and often segfaults.

The error handling is described in more detail in the "Error handling"
section of https://www.sqlite.org/backup.html.


mattn:
Thank you

PR #347, opened 2016-11-02
Add Go API for virtual tables

conorbranagan:
See https://www.sqlite.org/vtab.html for more details.

This work was started from
https://github.com/gwenn/gosqlite/blob/master/vtab.{c,go} and adds:

- Porting the API to go-sqlite3 APIs.
- Support for >= Go 1.6 without requiring the `cgocheck` flag to be changed.
- Filling out the unfinished callback functions for the `Vtable` struct.
- A simple `Context` API layer for ease of use when adding modules.

Tests are included.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,103 @@
+// Copyright (C) 2014 Yasuhiro Matsumoto <mattn.jp@gmail.com>.
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package sqlite3
+
+/*
+
+#ifndef USE_LIBSQLITE3
+#include <sqlite3-binding.h>
+#else
+#include <sqlite3.h>
+#endif
+#include <stdlib.h>
+// These wrappers are necessary because SQLITE_TRANSIENT
+// is a pointer constant, and cgo doesn't translate them correctly.
+
+static inline void my_result_text(sqlite3_context *ctx, char *p, int np) {
+	sqlite3_result_text(ctx, p, np, SQLITE_TRANSIENT);
+}
+
+static inline void my_result_blob(sqlite3_context *ctx, void *p, int np) {
+	sqlite3_result_blob(ctx, p, np, SQLITE_TRANSIENT);
+}
+*/
+import "C"
+
+import (
+	"math"
+	"reflect"
+	"unsafe"
+)
+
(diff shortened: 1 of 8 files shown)


mattn:
@gwenn

This change seems to be based on your's. If you allow, I'll merge this into go-sqlite3. How do you think? If you don't want, I'll close this.

gwenn:
@mattn No problem.
And currently, I don't have time but maybe go-sqlite3 users would be interested by this:
http://sqlite.org/carray.html
This function is the best replacement to the [intarray](https://github.com/gwenn/gosqlite/blob/master/intarray.go) virtual table.
Image that you have something like
`SELECT * FROM onetable WHERE onecolumn in (?, ?, ...)`
where the number of parameters is dynamic or large.
Currently it is a pain to handle this case.
But with the `carray` function, it becomes a lot easier.

conorbranagan:
Thanks @gwenn and thanks @mattn for taking a look. I realize this is a large change but I'm happy to make any necessary changes before getting it merged in.

mattn:
@gwenn thanks. @conorbranagan okay if you ready, please ping me again.

conorbranagan:
@mattn Sorry for the delay on getting back to this. I've added an additional example matching the existing GitHub repo virtual table example but using the pure Go API. This and the existing changes are good to go.

context.go
```diff
@@ -0,0 +1,103 @@
...
+	"unsafe"
+)
+
+const i64 = unsafe.Sizeof(int(0)) > 4
+
+type ZeroBlobLength int32
+type Context C.sqlite3_context
+
+// ResultBool sets the result of an SQL function.
+func (c *Context) ResultBool(b bool) {
```
mattn: In the feature, I have to add codes for context.Context. Could you please rename this to SQLiteContext?

conorbranagan: Good point, didn't think of this conflict at the time. Code is updated now. Let me know if there's anything else that needs fixing.

mattn:
LGTM

mattn:
Thank you very much.

gwenn:
I may be wrong but the implementation seems to leak memory.
Virtual table instances and Virtual table cursors will not be GCed until the connection is closed.
If the connection is pooled, all these instances created are not GCed when the connection is released to the pool.
Regards.

Issue #346, opened 2016-11-01
go1.8: support for additional features


kardianos:
Many new important features will land in go1.8. Please see document:
https://docs.google.com/document/d/1F778e7ZSNiSmbju3jsEWzShcb8lIO4kDyfKDNm4PNd8/edit?usp=sharing for the full list.

Because these features are new we want drivers to implement them before they are released. Ideally they should be implemented by 2016-12-12 to allow for any driver provided feedback from implementation if necessary.

mattn:
Look into it in later. But context.Context is used for the new feature. How do you think backward compatibility?

kardianos:
You make a valid point. I'm looking into it now.

kardianos:
Use "golang.org/x/net/context" and arrange your files like the sample in the google docs in the section "Supporting pre-go1.8 side by side."

mattn:
Okay, at the first, I'll add code without context features. ex ColumnTypeXXX. Thanks.

mattn:
I'm implementing HasNextResultSet, I have one question. When the several queries in a string, driver should send all of queries? if the part of query was wrong, following queries should not be executed?

kardianos:
My understand with sqlite (which could be wrong) is that it will execute
multiple sql statements in a single prepared batch so long as they are
separated by semicolon. However from initial looking it doesn't seem like
the sqlite engine supports multiple result sets. This is distinct from say
MS SQL Server or My SQL which will return from the server multiple result
sets from just a single query.

So I think you should just send any query string to sqlite like normally.
It just may not support NextResultSet, but unsure.

mattn:
Merged into master branch. https://github.com/mattn/go-sqlite3/pull/348

Thanks.

mattn:
However, I have some projects to do same. :-(

mattn:
Ah, I miss IsolationContext.

mattn:
SQLite3 doesn't support transaction isolate level. It only support single lock transaction per database. So I don't support IsolationContext for go-sqlite3.

@kardianos thanks your advices.

PR #348, opened 2016-11-04
new go1.8 feature

mattn:
Closing https://github.com/mattn/go-sqlite3/issues/346

## On Conn
* [x] BeginContext(ctx) (tx,err) found in ConnBeginContext
* [x] PrepareContext(ctx, query) (stmt,err) found in ConnPrepareContext
* [x] ExecContext(ctx, query, []namedvalue) (res, err)  found in ExecerContext
* [x] Ping(ctx) err found in Pinger
* [x] QueryContext(ctx, query, []namedvalue) (rows, err) found in QueryerContext

## On "Stmt"
* [x] ExecContext(ctx, []namedvalue) (res, err) found in StmtExecContext
* [x] QueryContext(ctx, []namedvalue) (rows, err) found in StmtQueryContext

## On "Rows"
* [ ] HasNextResultSet() bool found in RowsNextResultSet
* [ ] NextResultSet() error found in RowsNextResultSet
* [x] ColumnTypeDatabaseTypeName(index int) (name string)
* [x] ColumnTypeLength(index int) (length int64, ok bool)
* [x] ColumnTypeNullable(index int) (nullable, ok bool)
* [x] ColumnTypePrecisionScale(index int) (precision, scale int64, ok bool)
* [x] ColumnTypeScanType(index int) reflect.Type


mattn:
I decide NOT to support NextResultSet/HasNextResultSet because the go's multiple result is not same as sqlite's one.

azavorotnii:
@mattn is there some workaround for NextResultSet usage? Currently, if I have multiple selects in same Query() it returns only last select rows.

mattn:
When I looked at database / sql before, it seemed impossible to implement multiple-resultset for sqlite3 because the behavior of multiple-resultset of database/sql is different from the one expected. But now I looked it, I seems possible to implement. (Seems HasNextResultSet is gone, but not problem)

mattn:
https://github.com/mattn/go-sqlite3/issues/497

PR #351, opened 2016-11-07
Fix data race introduced on 605d9d08

andrefsp:
This commit will fix a data race introduced on 605d9d08 and also
make sure travis CI will now check for data races.


mattn:
make sense

mattn:
thank you

Issue #352, opened 2016-11-07
Building fails with "trace" tag


fancycode:
I'm trying to build e5a3c16c5c1d80b24f633e68aecd6b0702786d3d with `--tags "icu trace"` on `go version go1.6.2 linux/amd64` but got the following errors (the `icu` is most likely unrelated):
```
# github.com/mattn/go-sqlite3
could not determine kind of name for C._sqlite3_create_function
could not determine kind of name for C.uintptr_t
```

Ok, so it's the wrong name and a missing import:
```diff
diff --git a/tracecallback.go b/tracecallback.go
--- a/tracecallback.go
+++ b/tracecallback.go
@@ -13,6 +13,7 @@ package sqlite3
 #else
 #include <sqlite3.h>
 #endif
+#include <stdint.h>
 #include <stdlib.h>
 
 void stepTrampoline(sqlite3_context*, int, sqlite3_value**);
@@ -358,7 +359,7 @@ func (c *SQLiteConn) RegisterAggregator(name string, impl interface{}, pure bool
        if pure {
                opts |= C.SQLITE_DETERMINISTIC
        }
-       rv := C._sqlite3_create_function(c.db, cname, C.int(stepNArgs), C.int(opts), C.uintptr_t(newHandle(c, &ai)), nil, (*[0]byte)(unsafe.Pointer(C.stepTrampoline)), (*[0]byte)(unsafe.Pointer(C.doneTrampoline)))
+       rv := C.sqlite3_create_function(c.db, cname, C.int(stepNArgs), C.int(opts), C.uintptr_t(newHandle(c, &ai)), nil, (*[0]byte)(unsafe.Pointer(C.stepTrampoline)), (*[0]byte)(unsafe.Pointer(C.doneTrampoline)))
        if rv != C.SQLITE_OK {
                return c.lastError()
        }
```

But then I'm getting these:
```
# github.com/mattn/go-sqlite3
In file included from $WORK/github.com/mattn/go-sqlite3/_obj/_cgo_export.c:2:0:
vendor/src/github.com/mattn/go-sqlite3/tracecallback.go:72:13: error: conflicting types for ‘stepTrampoline’
 // for SQL Trace Hook = sqlite3_trace_v2():
             ^
In file included from $WORK/github.com/mattn/go-sqlite3/_obj/_cgo_export.c:2:0:
[...]

PR #353, opened 2016-11-08
Test different build configurations on Travis.

fancycode:
This helps to make sure that regressions like #352 are prevented.


mattn:
Good catch.

Issue #372, opened 2017-01-12
Cross compiling failed.


Cospotato:
### go Version
`go version go1.7.4 darwin/amd64`
### go env
```
GOARCH="amd64"
GOBIN="/Users/CosPotato/GoProjects/bin"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/CosPotato/GoProjects"
GORACE=""
GOROOT="/usr/local/go"
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
CC="clang"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/2g/4x1xzstj4vdbj939flg1q7v00000gn/T/go-build152787034=/tmp/go-build -gno-record-gcc-switches -fno-common"
CXX="clang++"
CGO_ENABLED="1"
```
### what i do
`GOOS=linux GOARCH=amd64 ENABLE_CGO=1 go build -o task-server task-server`
### what i get
```
github.com/mattn/go-sqlite3
GoProjects/src/github.com/mattn/go-sqlite3/tracecallback_noimpl.go:39: undefined: Error
```

mattn:
please update go-sqlite3

mattn:
BTW, should be CGO_ENABLED=1 not ENABLE_CGO=1

Cospotato:
Oh, sorry for the stupid mistake. I already update the go-sqlite3 with
`go get -u github.com/mattn/go-sqlite3`.
And then 
`GOOS=linux GOARCH=amd64 CGO_ENABLED=1 go build -o task-server task-server`.
Now the error is
```
# runtime/cgo
ld: unknown option: --build-id=none
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

mattn:
use gcc instead of clang.

Cospotato:
The macOS was linked all gcc compiler to clang. It is a huge project to remove llvm and install gcc. Maybe I should forget cross compile and compile it on real linux machine ); Thanks you.

mattn:
No, don't use host compiler. use cross-compiler.

mattn:
Ah, sorry. you already mentioned.

> Maybe I should forget cross c

markbates:
[...]

PR #373, opened 2017-01-18
removed two functions that aren't implemented that are causing build errors

markbates:
https://github.com/mattn/go-sqlite3/issues/372


mattn:
Maybe, you can remove the file.

mattn:
Thanks.

Issue #374, opened 2017-01-18
no buildable Go source files


markbates:
This is related to issue #372 

On a Mac, if I run `go build` everything works smoothly. If I run `GOOS=linux go build` I get the following:

```
go build github.com/mattn/go-sqlite3: no buildable Go source files in $GOPATH/src/github.com/mattn/go-sqlite3
```

mattn:
so closable?

PR #375, opened 2017-01-18
Removing an unused C import to allow for "buildable" go files. Fixes …

markbates:
…https://github.com/mattn/go-sqlite3/issues/374


mattn:
thank you

PR #376, opened 2017-01-26
add a FAQ entry about :memory: races

### LINKED ISSUE #204: "table not found" errors when concurrently reading from an in memory db
I ran into a bug where occasionally I was getting "table not found" errors for a database we create on startup and never write to or modify. I managed to reproduce it in the following gist

https://gist.github.com/jmunson/7b1974215d34439688b0 (name it main_test.go in an empty dir and go test -v should work)

This test file will create an in memory db and create a key->value table with 3 entries,  then we spawn 20 goroutines that each query for all 3 entries.

If this is ran with GOMAXPROCS set to 2 or more without setting MaxOpenConns to 1, then we get "table not found" errors.

I'm not sure if this is something that can be fixed or should just be documented around, but I thought I'd share in case anyone else runs into this.

### LINKED ISSUE #309: Document :memory: database connection issues, perhaps enforce a single connection.
I encountered a strange bug with an in-memory database where I was accidentally leaving transactions uncommitted.  To my surprise, it turns out that every new connection creates a brand-spanking new in-memory database, which meant that I wasn't running queries against the database I thought I was, but a blank database, resulting in some perplexing errors involving missing data that I was sure was actually there.

I eventually fixed the bug by eliminating an extraneous Begin() that had crept into my code, but I can't imagine a situation where "please run this query against a fresh database" would be desirable.  Could this behavior at the very least be documented somewhere?  It'd be nice if perhaps the driver could somehow limit the number of concurrent connections to in-memory databases to one, but you might still need to document this behavior so people don't get confused when they run into a deadlock caused by connection that was left open.

tych0:
Closes #204
Closes #309

Signed-off-by: Tycho Andersen <tycho@tycho.ws>

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,15 @@ FAQ
 
     Yes for readonly. But, No for writable. See #50, #51, #209.
 
+* Why is it racy if I use a `sql.Open("sqlite", ":memory:")` database?
+
+    Each connection to :memory: opens a brand new in-memory sql database, so if
+    the stdlib's sql engine happens to open another connection and you've only
+    specified ":memory:", that connection will see a brand new database. A
+    workaround is to use "file::memory:?mode=memory&cache=shared". Every
+    connection to this string will point to the same in-memory database. See
+    #204 for more info.
+
 License
 -------
 


mattn:
Great. Thanks

PR #398, opened 2017-03-24
Removed ambitious conn.Close()

mkungla:


mattn:
Sorry & Thank you.

mkungla:
np

gaillard:
There might not be a conn var at this point but isn't the connection still open at this point and need to be closed? The _sqlite3_open_v2 has still been called.

PR #399, opened 2017-03-24
Adding hook to vtable to allow VTab's to be updated

kenshaw:
This commit changes the vtable 'xUpdate' goModule field to a new
'cXUpdate' callback function which in turns calls a 'goVUpdate'
callback.

This new callback allows Go defined virtual table implementations
satisfying the VTabUpdater interface (also newly defined) a way to
delete/insert/update rows in a VTab.

Additionally, an anonymous interface is used within the goVUpdate
callback looking for 'TableName() string' which, when defined on a VTab
is used to provide a better contextual error message to end users if the
VTab is read only.

Care was taken to follow existing code style/conventions for this
addition, and for backwards-compatibility with existing VTab
implementations (hence why a new interface was required).

### DIFF
--- a/sqlite3_vtable.go
+++ b/sqlite3_vtable.go
@@ -182,6 +182,19 @@ static int cXRowid(sqlite3_vtab_cursor *pCursor, sqlite3_int64 *pRowid) {
 	return SQLITE_OK;
 }
 
+char* goVUpdate(void *pVTab, int argc, sqlite3_value **argv, sqlite3_int64 *pRowid);
+
+static int cXUpdate(sqlite3_vtab *pVTab, int argc, sqlite3_value **argv, sqlite3_int64 *pRowid) {
+	char *pzErr = goVUpdate(((goVTab*)pVTab)->vTab, argc, argv, pRowid);
+	if (pzErr) {
+		if (pVTab->zErrMsg)
+			sqlite3_free(pVTab->zErrMsg);
+		pVTab->zErrMsg = pzErr;
+		return SQLITE_ERROR;
+	}
+	return SQLITE_OK;
+}
+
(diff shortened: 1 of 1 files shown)


kenshaw:
I have posted this as a WIP, as I realize that there is likely more work that would need to be done to be accepted, such as unit tests, etc. However, I have gone ahead and put this up in the hopes of determining if this would be accepted in its general design, to see if the API design is satisfactory/unsatisfactory, and if so, to start discussion on what, if any API/interface would be acceptable instead.

Appreciate the consideration. Thanks in advance!

kenshaw:
Any feedback on this PR? Let me know if this (or some kind of alternate form) would be a viable addition! Would love to get this in, somehow, into the main code base.

For reference, I'm working on a separate Go-based CSV driver, that uses this amazing package, and would like it to be able to "update" and "insert" rows. That can't be done without the update hooks. 

Similarly, I had the idea to do the same for Google Sheets. I know it sounds silly, but I just get these ideas in my head and would love to be able to get them down. I'm more than happy to pursue these separately, but I really believe the update functionality directly in this package makes the most sense.

mattn:
Sorry my delay, I'll look into it in later.

mattn:
LGTM

mattn:
Ah, test not found. could you please add test?

kenshaw:
Great, thanks, appreciate it! Yes, I'll add a test immediately. I originally had just posted this in order to get feedback if the approach would be accepted or not, as the PR introduced a new type.

kenshaw:
Sorry for the delay, I should have this done today.

mattn:
NP. I don't hurry. Feel free to do it. :)

Issue #255, opened 2015-11-19
howto make pragma statements persistent, e.g. for foreign_keys checks


bernhardreiter:
Hello!

Writing a small RESTful webserver application, we rely on constrains in the sqlite db.
In our app the foreign_key checks were working once every second time. We figured that
go-sqlite3 will open several connections, probably recycling some, and some of the pool did not have
the "PRAGMA foreign_keys = ON" set.

Our solution was to use a connect_hook like

``` go
import (
        "database/sql"

        sqlite3 "github.com/mattn/go-sqlite3"
)

func init() {
        sql.Register("sqlite3_with_fk",
                &sqlite3.SQLiteDriver{
                        ConnectHook: func(conn *sqlite3.SQLiteConn) error {
                                _, err := conn.Exec("PRAGMA foreign_keys = ON", nil)
                                return err
                        },
                })
}
```

and later calling

``` go
sql.Open("sqlite3_with_fk", dataSourceName)
```

Is this the best way of doing this?

Because this is a common use case, people wanting to set some sqlite PRAGMAs
and use them everytims, we suggest to add a hint to the documentation.

Best Regards and thanks for creating go-sqlite3 as Free Software!
Bernhard

gwenn:
http://sqlite.org/compile.html#default_foreign_keys
Regards.

bernhardreiter:
On Thursday 19 November 2015 at 18:23:30, gwenn wrote:
> http://sqlite.org/compile.html#default_foreign_keys

Hi gwenn,
I'm assuming that you are are pointing me to the possibilty to change
the default for sqlite at build time. Thanks for the hint!

Yes, I knew about this possibility, that it sometimes is not a good
option because an already build sqlite version is uses. Like if you
are on a Debian GNU/Linux you really would want to use the system sqlite
builds. One argument for this are security updates.

Therefore I think this will be common use case which should be solved on 
go-sqlite3 level. And with the connect hook, there is a reasonable solution.
The behaviour of a go application with go-sqlite3 can be surprising, at least 
it took us several hours to figure out what was going on and find a good 
solution. A hint in the go-sqlite3 would have been very helpful. Thus we are 
suggesting it.

Regards,
Bernhard

gwenn:
There is another alternative:
http://sqlite.org/c3ref/open.html#urifilenamesinsqlite3open
http://sqlite.org/c3ref/uri_boolean.html
 
```go
[...]

PR #407, opened 2017-04-01
Add _foreign_keys connection parameter

### LINKED ISSUE #377: Foreign key constraint just doesn't work
I am probably misunderstanding something, but it appears that foreign key constraints are ignored no matter what I do. Here's an example:

````
package main

import (
	"database/sql"

)

func main() {
	db, _ := sql.Open("sqlite3", "test.db")
	tx, _ := db.Begin()
	defer tx.Commit()
	tx.Exec("PRAGMA foreign_keys = ON")
	tx.Exec("create table foo (a)")
	tx.Exec("create table bar (a references foo(a))")
	_, err := tx.Exec("insert into bar values ('lol')")
	if err != nil {
		panic(err.Error())
	}
}
````

Expected result: panic. Actual result: `test.db` gets created, with the values being inconsistent with the foreign key constraint.

Why is this happening? I even recall using this package before with foreign keys without issue...

zombiezen:
Fixes #377
Updates #255

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -400,14 +400,18 @@ func (c *SQLiteConn) AutoCommit() bool {
 }
 
 func (c *SQLiteConn) lastError() error {
-	rv := C.sqlite3_errcode(c.db)
+	return lastError(c.db)
+}
+
+func lastError(db *C.sqlite3) error {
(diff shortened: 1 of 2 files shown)


mattn:
This keep default behavior, good.

Thank you.

PR #408, opened 2017-04-01
use for Go 1.8 "context" standard package, not "golang.org/x/net/cont…

bstarynk:
Go 1.8 (& probably Go 1.7) has a standard internal `"context"` package which should be imported (in  file `sqlite3.go` in place of `"golang.org/x/net/context"` external package.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -116,7 +116,10 @@ import (
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


mattn:
This break go1.6 .

bestplay:
I use this `go tool fix -force context` to upgrading the import path of Context package.

PR #409, opened 2017-04-07
Adding unit test for VTable Insert/Update/Delete

kenshaw:
This fixes a small piece of logic in the original change, as well as adds an additional unit test for the VTable Insert/Update/Delete func's.

Test with:
```sh
$ go test -tags vtable -v -x -run TestVUpdate
```

### DIFF
--- a/sqlite3_vtable_test.go
+++ b/sqlite3_vtable_test.go
@@ -8,8 +8,11 @@ package sqlite3
 
 import (
 	"database/sql"
+	"errors"
 	"fmt"
 	"os"
+	"reflect"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


kenshaw:
@mattn Apology on the delay in finishing this, but had other work to attend to.

sqlite3_vtable_test.go
```diff
@@ -148,8 +151,335 @@ func TestCreateModule(t *testing.T) {
...
+	}
+	i, err := getRowCount(rows)
+	if err != nil {
+		t.Fatalf("expected no error, got: %v", err)
+	}
+	if i != 0 {
+		t.Fatalf("expected no rows in vt, got: %d", i)
+	}
+
+	_, err = db.Exec(`delete from vt where f1 = 'yes'`)
```
mattn: Where is INSERT for this?

mattn:
totally okay to me (except last question)

sqlite3_vtable_test.go
```diff
@@ -148,8 +151,335 @@ func TestCreateModule(t *testing.T) {
...
+	}
+
+	// test bad column name
+	_, err = db.Exec(`insert into vt (f4) values('a')`)
+	if err == nil {
+		t.Fatalf("expected error on insert, got nil")
+	}
+
+	// insert to vt
+	res, err := db.Exec(`insert into vt (f1, f2, f3) values (115, 'b', 'c'), (116, 'd', 'e')`)
```
kenshaw: @mattn the insert is here

sqlite3_vtable_test.go
kenshaw: line 220 -- the delete is there to check that a delete still works when the table is empty. It doesn't have much to do with the actual vUpdate stuff, but I get carried away when writing unit tests to make sure start conditions are always the same.

kenshaw: I completely understand that this specific check doesn't add much, but there's a lot associated with the module implementation that is the actual logic. I'm happy to remove this part if you think that would be better.

sqlite3_vtable_test.go
mattn: right.

mattn:
Thank you very much.

PR #415, opened 2017-05-03
Add time format with "AM/PM"

ptsolmyr:
While using this sqlite3 driver, we found it couldn't process time format with "am/pm" correctly.


mattn:
sqlite3 doesn't handle time format strictly. So you can store time in any formats. I don't want to add code each times for someone want to add new format.

GJRTimmer:
@mattn Please close.

PR #423, opened 2017-06-09
Add support for collation sequences implemented in Go.

danderson:
This allows Go programs to register custom comparison functions with
sqlite, and ORDER BY that comparator.


danderson:
The Travis failures look unrelated to this change, there's a new data race in transaction handling with Go tip.

mattn:
Thank you

Issue #424, opened 2017-06-14
Empty byte array is not NULL


xxr3376:
`[]byte{}` should be converted to empty BLOB instead of `NULL`, `[]byte{} != nil` in go.

I use following table to reproduce my issue.

```sql
CREATE TABLE test (
    field BLOB NOT NULL
);
```
Following code will report `NOT NULL constraint failed` error, but it's ok to execute `INSERT INTO test (field) VALUES ('')` in command line tool.

```go
_, err := db.Exec(`INSERT INTO test (field) VALUES (?)`, []byte{})
// err = "NOT NULL constraint failed"
```

https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L776 calls `sqlite3_bind_blob` with buffer=nil when len(buffer) = 0, but according to [Sqlite C Interface](https://sqlite.org/c3ref/bind_blob.html), it shouldn't.

> If the third parameter to sqlite3_bind_text() or sqlite3_bind_text16() or sqlite3_bind_blob() is a NULL pointer then the fourth parameter is ignored and the end result is the same as sqlite3_bind_null().

Maybe we should `make([]byte, 1)` then pass its address and length=0 to `sqlite3_bind_blob`.

mattn:
Could you please send me pull-request?

PR #425, opened 2017-06-14
Treat []byte{} as empty BLOB instead of NULL.

xxr3376:
This resolve issue https://github.com/mattn/go-sqlite3/issues/424 .


mattn:
Thank you

Issue #426, opened 2017-06-15
Data race during database close


otoolep:
I use this project in [rqlite](https://github.com/rqlite/rqlite). My unit testing seems has hit a data race I haven't seen before. This may be something I need to handle on my side, but I am surprised it's never shown up before in my testing. Any ideas?

https://circleci.com/gh/rqlite/rqlite/1158#tests/containers/1
```
WARNING: DATA RACE
Write at 0x00c4201d2c80 by goroutine 39:
  github.com/mattn/go-sqlite3.(*SQLiteConn).Close()
      /home/ubuntu/.go_workspace/src/github.com/mattn/go-sqlite3/sqlite3.go:682 +0x7b
  github.com/rqlite/rqlite/db.(*DB).Close()
      /home/ubuntu/.go_project/src/github.com/rqlite/rqlite/db/db.go:134 +0x4c
  github.com/rqlite/rqlite/store.(*Store).Close()
      /home/ubuntu/.go_project/src/github.com/rqlite/rqlite/store/store.go:280 +0x52
  github.com/rqlite/rqlite/system_test.(*Node).Deprovision()
      /home/ubuntu/.go_project/src/github.com/rqlite/rqlite/system_test/helpers.go:35 +0x57
  github.com/rqlite/rqlite/system_test.Test_MultiNodeCluster()
      /home/ubuntu/.go_project/src/github.com/rqlite/rqlite/system_test/cluster_test.go:147 +0xb0a
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:657 +0x107

Previous read at 0x00c4201d2c80 by goroutine 50:
  github.com/mattn/go-sqlite3.(*SQLiteStmt).Close()
      /home/ubuntu/.go_workspace/src/github.com/mattn/go-sqlite3/sqlite3.go:716 +0x132
  github.com/mattn/go-sqlite3.(*SQLiteConn).exec()
      /home/ubuntu/.go_workspace/src/github.com/mattn/go-sqlite3/sqlite3.go:456 +0x1f3
  github.com/mattn/go-sqlite3.(*SQLiteConn).Exec()
      /home/ubuntu/.go_workspace/src/github.com/mattn/go-sqlite3/sqlite3.go:427 +0x2a3
  github.com/rqlite/rqlite/db.(*DB).Execute.func1()
      /home/ubuntu/.go_project/src/github.com/rqlite/rqlite/db/db.go:243 +0x2cd
[...]

PR #427, opened 2017-06-17
Address data race during close database and statement

otoolep:


otoolep:
This PR should fix https://github.com/mattn/go-sqlite3/issues/426, but I don't think it's a comprehensive fix for the handling of `db` in this code.

mattn:
LGTM Thank you

Issue #428, opened 2017-06-20
Commit 308f5f1b2fae699cb38474136205cbab8f081019 - "Treat []byte{} as empty bytes instead of NULL." causes cgo argument has Go pointer to Go pointer panic 


emakeev:
Commit 308f5f1b2fae699cb38474136205cbab8f081019 reverted "bind: pass &v[0] in direct call to C" fix (commit b76c61051faaf0baf3215e6f811003d98639b702) and we started seeing the following cgo panic on some of our unit tests:

... runtime error: cgo argument has Go pointer to Go pointer

panic(0xacd980, 0xc42036eaf0)
	/usr/lib/go-1.8/src/runtime/panic.go:489 +0x2cf
github.com/mattn/go-sqlite3.(*SQLiteStmt).bind.func11(0x7f9680015a08, 0xc400000002, 0xc420403bb0, 0xc40000003c, 0x0)
	/home/vagrant/external/src/github.com/mattn/go-sqlite3/sqlite3.go:795 +0x8b
github.com/mattn/go-sqlite3.(*SQLiteStmt).bind(0xc4202fa030, 0xc42039caf0, 0x2, 0x2, 0x7f9680015a08, 0x7f9600000002)
	/home/vagrant/external/src/github.com/mattn/go-sqlite3/sqlite3.go:795 +0x3a3
github.com/mattn/go-sqlite3.(*SQLiteStmt).exec(0xc4202fa030, 0x7f9698b98890, 0xc420062e38, 0xc42039caf0, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0)
	/home/vagrant/external/src/github.com/mattn/go-sqlite3/sqlite3.go:872 +0x86
github.com/mattn/go-sqlite3.(*SQLiteConn).exec(0xc420060f00, 0x7f9698b98890, 0xc420062e38, 0xc42039c9b0, 0x50, 0xc42039caf0, 0x2, 0x2, 0x0, 0x0, ...)
	/home/vagrant/external/src/github.com/mattn/go-sqlite3/sqlite3.go:449 +0x2f9
github.com/mattn/go-sqlite3.(*SQLiteConn).ExecContext(0xc420060f00, 0x10b0340, 0xc420062e38, 0xc42039c9b0, 0x50, 0xc42039caa0, 0x2, 0x2, 0xc42045d800, 0xc4203dab60, ...)
[...]

PR #429, opened 2017-06-21
Fix for cgo panic, issue #428: https://github.com/mattn/go-sqlite3/is…

emakeev:
fix for issue #428, tested with go test go-sqlite3/...


mattn:
Thanks

PR #436, opened 2017-07-09
Add connection option for recursive triggers

zombiezen:
Similar to foreign keys, the recursive triggers PRAGMA affects the
interpretation of all statements on a connection.


mattn:
Thank you

PR #439, opened 2017-07-14
Add support for sqlite3_unlock_notify

typeless:
For https://github.com/go-gitea/gitea/issues/2040

Added new build tag `sqlite_unlock_notify`.

```
name      old req/s   new req/s   delta            
Exec       219k ±20%   217k ±18%     ~     (p=0.690 n=5+5)                                            
Query     75.8k ±12%  67.9k ±13%     ~     (p=0.095 n=5+5)                                            
Params    74.1k ± 3%  68.5k ± 6%   -7.63%  (p=0.008 n=5+5)                                            
Stmt       119k ± 1%   108k ± 5%   -9.56%  (p=0.008 n=5+5)                                            
Rows      2.21k ± 1%  2.08k ± 1%   -5.61%  (p=0.016 n=5+4)                                            
StmtRows  2.30k ± 2%  2.05k ±11%  -10.78%  (p=0.008 n=5+5)  
```

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -78,8 +78,38 @@ _sqlite3_exec(sqlite3* db, const char* pcmd, long long* rowid, long long* change
   return rv;
 }
 
+#ifdef SQLITE_ENABLE_UNLOCK_NOTIFY
+extern int sqlite3_step_blocking(sqlite3_stmt *stmt);
+extern int _sqlite3_step_blocking(sqlite3_stmt* stmt, long long* rowid, long long* changes);
+extern int sqlite3_prepare_v2_blocking(sqlite3 *db, const char *zSql, int nBytes, sqlite3_stmt **ppStmt, const char **pzTail);
+
 static int
-_sqlite3_step(sqlite3_stmt* stmt, long long* rowid, long long* changes)
+sqlite3_step_internal(sqlite3_stmt *stmt)
+{
+  return sqlite3_step_blocking(stmt);
+}
+
+static int
+_sqlite3_step_internal(sqlite3_stmt* stmt, long long* rowid, long long* changes)
+{
+  return _sqlite3_step_blocking(stmt, rowid, changes);
+}
+
+static int
+sqlite3_prepare_v2_internal(sqlite3 *db, const char *zSql, int nBytes, sqlite3_stmt **ppStmt, const char **pzTail)
+{
+  return sqlite3_prepare_v2_blocking(db, zSql, nBytes, ppStmt, pzTail);
+}
+
+#else
+static int
+sqlite3_step_internal(sqlite3_stmt *stmt)
+{
+  return sqlite3_step(stmt);
+}
+
+static int
+_sqlite3_step_internal(sqlite3_stmt* stmt, long long* rowid, long long* changes)
 {
   int rv = sqlite3_step(stmt);
   sqlite3* db = sqlite3_db_handle(stmt);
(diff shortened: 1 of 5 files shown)


mattn:
kick travis

mattn:
Could you please rebase ?

typeless:
I'd like to add a test for what the PR tries to fix but cannot deliver it in the first few attempts. I'll try later when I have more free time.

GJRTimmer:
@typeless Could you rebase

typeless:
@GJRTimmer Done

mattn:
I wonder this can be tested.

sqlite3.go
```diff
@@ -81,8 +82,25 @@ _sqlite3_exec(sqlite3* db, const char* pcmd, long long* rowid, long long* change
 static int
 _sqlite3_step(sqlite3_stmt* stmt, long long* rowid, long long* changes)
 {
-  int rv = sqlite3_step(stmt);
+  extern int unlock_notify_wait(sqlite3 *db);
+  int rv;
   sqlite3* db = sqlite3_db_handle(stmt);
+
+  for (;;) {
```
GJRTimmer: Shouldn't this be
```go
for {
```

GJRTimmer:
@mattn I think you are right we need some code to test it, maybe even an example.

@typeless Instead of currently trying to write a test for it, You did probably design this with some code to see if it worked. Could you add an example for this to the `_example` folder ?

sqlite3.go
typeless: I don't have access to the dev environment right now but this `for` looks like C code.

bvinc:
I notice that in the step function, you check the extended error code to make sure it's `SQLITE_LOCKED_SHAREDCACHE`, but you don't do this in the prepare code.  Why not?

typeless:
@bvinc It was modified from the example code:  https://sqlite.org/unlock_notify.html

The explanation of the `SQLITE_LOCKED_SHAREDCACHE` check is described in the last few paragraphs.
It seems to me that `sqlite3_blocking_prepare_v2` would not result in the same scenario as `qlite3_blocking_step` does.

typeless:
By the way, I have found a test which can reproduce the "database is locked" error but it turned out that the PR does not fix the problem. I'll try to figure it out later and see if this PR has any hope to be helpful.

bvinc:
I found your test case in a closed PR.  I assume this is the test case that you're talking about.  I believe that there is some confusion again between `SQLITE_BUSY` and `SQLITE_LOCKED`.

Without using shared cache mode (which applies to your test), when two transactions conflict, the one that fails uses the busy handler, which waits a small number of random milliseconds, and then tries again.  Once this fails for 5 seconds (which is the default busy timeout for this driver), then it fails with `SQLITE_BUSY`, which says `database is locked`.  What's happening with your test case is that one statement is spinning so fast that it's completely starving out the other statement for a full 5 seconds.  This is completely normal and just how SQLite works.  You can see that if you increase the busy timeout by adding "?_busy_timeout=60000" that it won't give an error anymore.

But *this* PR is about something completely different.  This PR is about stopping people from getting `SQLITE_LOCKED` when using shared cache mode (which your test case did not use).

typeless:
@bvinc You're correct I misunderstood the two different errors. I need to try it again when I come back to this.

typeless:
@mattn @GJRTimmer @bvinc  The test has been added. PTAL.

typeless:
Hmmm, I have to make it compatible with `USE_LIBSQLUTE3` .

typeless:
Found a bug while trying to integrate the change into Gitea. I am looking into it.

mattn:
Thank you. Sounds good to me. One thing. please add `_` for the functions always to be not use same namespace `sqlite_`.

typeless:
@mattn Done. Is that what you mean?

mattn:
Yes! Thank you.

PR #454, opened 2017-08-21
Fix to better handle NULL values in TEXT and BLOB columns.

### LINKED ISSUE #453: Recent bug with nil TEXT and BLOB columns
I made a quick reproduction of the bug. tl;dr: A null value in a text or blob column returns a []byte{0} presumably something wants to return an empty string in such cases now.

```go
package repro

import (
        "database/sql"
        "fmt"

)

func Example() {
        db, _ := sql.Open("sqlite3", "main.db")
        db.Exec("create table tbl (txt text)")
        var b []byte
        db.Exec("insert into tbl (txt) values (?)", &b)
        rows, _ := db.Query("select txt from tbl ")
        rows.Next()
        rows.Scan(&b)
        fmt.Printf("%#v\n", b)

        // Output:
        // []byte(nil)
}

// Actual:
// --- FAIL: Example (0.01s)
// got:
// []byte{0x0}
// want:
// []byte(nil)
// FAIL
// exit status 1
// FAIL repro   0.011s
```

gholt:
Fixes #453

### DIFF
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -6,6 +6,7 @@
 package sqlite3
 
 import (
+	"bytes"
 	"database/sql"
 	"database/sql/driver"
 	"errors"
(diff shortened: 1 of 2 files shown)


mattn:
Good catch. Thanks

sqlite3_test.go
```diff
@@ -1343,6 +1344,61 @@ func TestUpdateAndTransactionHooks(t *testing.T) {
...
+		}
+		var scanBytes []byte
+		if err = rows.Scan(&scanBytes); err != nil {
+			t.Fatal(tst.name, err)
+		}
+		if err = rows.Err(); err != nil {
+			t.Fatal(tst.name, err)
+		}
+		if tst.expectedBytes == nil && scanBytes != nil {
+			t.Errorf("%s: %#v != %#v", tst.name, scanBytes, tst.expectedBytes)
```
gwenn:
```go
		} else if tst.expectedBytes != nil && scanBytes == nil {
			t.Errorf("%s: %#v != %#v", tst.name, scanBytes, tst.expectedBytes)
```
http://sqlite.org/c3ref/column_blob.html
> The return value from sqlite3_column_blob() for a zero-length BLOB is a NULL pointer.

PR #462, opened 2017-09-06
Updated "context" import since it has become a standard library 

faruzzy:
After "go 1.7", "context" has become a standard library
Please see: https://golang.org/doc/go1.7#context


mattn:
Yes, known issue. This is hack for compatibility. I should decide dropping go1.6 .

at15:
@mattn is there anything block this PR from being merged?

mattn:
Now go-sqlite3 use context instead of x/net/context. Thank you.

PR #463, opened 2017-09-13
Allow NOMUTEX|READONLY Open Mode

fortytw2:
This PR is more for discussion - definitely don't merge in this current state.

I couldn't find a way to open a database in read only mode (so that no locks are ever acquired on the DB, essential for highly concurrent readers), so I've hacked this in.

Any thoughts on how this interface should look for the end users, @mattn?


mattn:
You can specify DSN like below.

```
db, err := sql.Open("sqlite3", file:test.db?immutable=1")
```

fortytw2:
Ah, excellent. I was rather confused as to where this was set (L40-45 of sqlite3.go) and to how it interacted with the flags to sqlite_open_v2. Thanks!

Issue #190, opened 2015-03-25
'sql: Scan error on column index 0: unsupported driver -> Scan pair: []uint8 -> *time.Time' with datetime('now')


bbigras:
[sqlite3_column_decltype()](https://www.sqlite.org/c3ref/column_decltype.html) returns NULL when it's an expression.

I'm not sure it's a bug.

I was thinking that maybe go-sqlite3 could guess that the string is a datetime but that could cause problem when someone want a text timestamp on purpose.

It could be easier to do the work on the user side since we know if we expect a string or a datetime/date/timestamp.

``` go
func TestDateTimeNow(t *testing.T) {
    tempFilename := TempFilename()
    db, err := sql.Open("sqlite3", tempFilename)
    if err != nil {
        t.Fatal("Failed to open database:", err)
    }
    defer db.Close()

    var d time.Time
    err = db.QueryRow("select datetime('now')").Scan(&d)
    if err != nil {
        t.Fatal("Failed to scan datetime:", err)
    }
}
```

```
--- FAIL: TestDateTimeNow (0.02s)
        sqlite3_test.go:917: Failed to scan datetime: sql: Scan error on column index 0: unsupported driver -> Scan pair: []uint8 -> *time.Time
FAIL
```

mattn:
`datetime('now')` has no type affinity.

http://www.sqlite.org/datatype3.html#expraff

mattn:
So you need to use `cast(datetime('now') as text)`

brunoqc:
I still have the problem if I cast to text.

```go
func TestDateTimeNow(t *testing.T) {
	tempFilename := TempFilename()
	db, err := sql.Open("sqlite3", tempFilename)
	if err != nil {
		t.Fatal("Failed to open database:", err)
	}
	defer db.Close()

	var d time.Time
	err = db.QueryRow("SELECT cast(datetime('now') as text)").Scan(&d)
	if err != nil {
		t.Fatal("Failed to scan datetime:", err)
	}
}
```

```
--- FAIL: TestDateTimeNow (0.01s)
        sqlite3_test.go:960: Failed to scan datetime: sql: Scan error on column index 0: unsupported driver -> Scan pair: []uint8 -> *time.Time
```

It seems the type was already text.

```
sqlite> SELECT typeof(datetime('now'));
text
sqlite> SELECT typeof(cast(datetime('now') as text));
text
```

gwenn:
[...]

Issue #316, opened 2016-07-08
Can't scan datetime into time.Time


infogulch:
Trying to scan a *time.Time (`select datetime('now')`) using the below running example gives the error:

> panic: cannot scan row: sql: Scan error on column index 0: unsupported Scan, storing driver.Value type []uint8 into type *time.Time

I would expect it to not give an error and instead populate the Time variable with the result, because time.Time/datetime is [in the list of supported types](https://godoc.org/github.com/mattn/go-sqlite3#hdr-Supported_Types).

However, changing the Time variable to string works as expected and outputs: 

> the time is: 2016-07-08 22:01:38

```go
package main

import (
	"database/sql"
	"fmt"
	"time"

	_ "github.com/mattn/go-sqlite3"
)

func main() {
	db, err := sql.Open("sqlite3", ":memory:")
	if err != nil {
		panic(fmt.Sprintf("cannot open db: %v", err))
	}
	err = db.Ping()
	if err != nil {
		panic(fmt.Sprintf("cannot open db: %v", err))
	}
	// interesting stuff starts here
	row := db.QueryRow(`select datetime('now')`)
	var t time.Time
	// var t string  // NOTE: using this declaration for t works
	if err = row.Scan(&t); err != nil {
		panic(fmt.Sprintf("cannot scan row: %v", err))
	}
	fmt.Printf("the time is: %s\n", t)
}
```

gwenn:
See issue #190

infogulch:
Thanks for the link. Closing as a duplicate.

PR #468, opened 2017-10-16
add a case in Next scan to fix issue #190 and issue #316

gingerhot:
I get a same error as in #190 and #316 when I  select a nullable datetime typed column `created_at` in one table:

```sql
SELECT created_at FROM users;
```
then  I use `COALESCE` and `CAST` to give it a default value and cast it as text:

```sql
SELECT cast(coalesce(created_at, '2017-10-16T08:23:19.120Z') as text) as created_at FROM users;
```
and I get the same error.

So I add a case to the `Next()` scan to give datetime typed column a chance to be parsed. After that issue #190/#316 and my problem are all fixed. 

I'm not sure if this's the right approach to the problem, here's twice benchmark results run on my computer:

1. current master version's:
```bash
 ~/go/src/github.com/mattn/go-sqlite3 (master) $ go test
BenchmarkExec           200000     151258 req/s
BenchmarkQuery          100000      60628 req/s
BenchmarkParams         100000      55500 req/s
BenchmarkStmt           100000      74674 req/s
BenchmarkRows             2000       1995 req/s
BenchmarkStmtRows         3000       2041 req/s
PASS
ok  	github.com/mattn/go-sqlite3	22.608s
 ~/go/src/github.com/mattn/go-sqlite3 (master) $ go test
BenchmarkExec           200000     134008 req/s
BenchmarkQuery          100000      55242 req/s
BenchmarkParams         100000      56315 req/s
BenchmarkStmt           100000      74565 req/s
BenchmarkRows             3000       2034 req/s
BenchmarkStmtRows         3000       2033 req/s
PASS
ok  	github.com/mattn/go-sqlite3	23.432s
```

2. the version of this pull request:

```bash
~/go/src/github.com/goonr/go-sqlite3 (master) $ go test
BenchmarkExec           200000     127702 req/s
BenchmarkQuery          100000      57204 req/s
BenchmarkParams         100000      48265 req/s
BenchmarkStmt           100000      73005 req/s
BenchmarkRows             3000       1994 req/s
BenchmarkStmtRows         3000       2034 req/s
PASS
ok  	github.com/goonr/go-sqlite3	23.628s
~/go/src/github.com/goonr/go-sqlite3 (master) $ go test
BenchmarkExec           200000     147327 req/s
BenchmarkQuery          100000      52995 req/s
BenchmarkParams         100000      42885 req/s
BenchmarkStmt           100000      64088 req/s
BenchmarkRows             2000       1990 req/s
BenchmarkStmtRows         3000       2054 req/s
PASS
ok  	github.com/goonr/go-sqlite3	23.691s
```

Thanks~

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1150,6 +1150,27 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 					t = t.In(rc.s.c.loc)
 				}
 				dest[i] = t
+				// this case is for a nullable datetime typed colomn a chance to be parsed
+				// to resolve the problem in issue #190 and issue #316
+			case "":
+				var t time.Time
+				s = strings.TrimSuffix(s, "Z")
+				for _, format := range SQLiteTimestampFormats {
+					if timeVal, err = time.ParseInLocation(format, s, time.UTC); err == nil {
+						t = timeVal
+						break
+					}
+				}
+				// if not parsed as a datetime type successfully, do as the default case does
+				if err != nil {
+					dest[i] = []byte(s)
+				} else {
+					if rc.s.c.loc != nil {
+						t = t.In(rc.s.c.loc)
+					}
+					dest[i] = t
+				}
+
 			default:
 				dest[i] = []byte(s)
 			}
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -1111,7 +1111,7 @@ func TestDateTimeNow(t *testing.T) {
 	defer db.Close()
 
 	var d time.Time
-	err = db.QueryRow("SELECT datetime('now')").Scan(TimeStamp{&d})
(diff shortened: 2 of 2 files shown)


mattn:
This breaks compatibility. And when the text is likely date format (but not datetime), it will be confusing.

gingerhot:
About the compatibility, I think maybe or maybe not. Because when you scan a struct that has a nullable timestamp attribute directly, there will be an error as same as in #190, you must do some trick, so now you can remove those tricks.

And for the date format likely text, it would really make confusing.  But maybe the user can avoid it by defining a such a column as a not NULL field. Anyway, I think it's a problem.

mattn:
Sorry, I'll close this since type of `datetime('now')` is string and I can not fix this without hack. If you want to scan it as time.Time, Please use sql.Scanner. https://github.com/mattn/go-sqlite3/issues/190#issuecomment-343341834

Issue #472, opened 2017-10-25
$N style placeholders do not work for selects if they are not in ascending order.


samuelkaufman:
## Description

When using SQL generating software, the bind params are not always in ascending order. I'd like to put in $x in the right spot, and have it bind properly,
with a case like:

    row = db.QueryRow(`select id, name from foo where id = $2 and name = $1`, text, 1)

## Expected

id would get the second param, 1, and name would get the first param, text.

## Actual

id gets the text param, name gets the 1 param.

samuelkaufman:
See also https://github.com/mattn/go-sqlite3/issues/187

mattn:
SQLite3 doesn't support $N style placeholder. Please use :N instead.

samuelkaufman:
OK thanks! I couldn't find any docs but I saw tests that used the $N syntax and it sort of worked until i had sql with $2 and $1 reversed.

mattn:
certainly, go-sqlite3 had supported $N style because older Go didn't have sql.NamedParam. But Go18 added it. So I removed support of $N style.

mattn:
As I mentioned in above, $N was no longer supported.

jrefior:
I wonder if this could be documented somewhere? At least I didn't find anything about placeholders in godoc or the README.

I just spent about 3 hours trying to figure out why a table update seemed to be silently failing with this driver, turned out it was placeholder order, when my query had $2 then $1.

I wish I'd thought to search sqlite documentation on placeholders before, I'm used to postgresql. It looks like this is the relevant section, correct? https://www.sqlite.org/lang_expr.html

Should all of those work, or what is the current recommended placeholder style? I see examples/simple.go uses ? as placeholders.

Thanks for the library, I haven't run into other issues.

PR #473, opened 2017-10-25
added a failing test "TestBindOrder" which demonstrates that this dri…

samuelkaufman:
…ver cannot handle $1, $2,type

bind params when they are out of order,
ex: `SELECT blah from foo where bar=$2 and foo=$1`

See issue https://github.com/mattn/go-sqlite3/issues/472


mattn:
SQLite3 doesn't support $N.

https://www.sqlite.org/c3ref/bind_blob.html

PR #479, opened 2017-11-05
Move RegisterAggregator implementation

kenshaw:
The SQLiteConn.RegisterAggregator implementation was defined in
sqlite3_trace.go file, which is guarded with a build constraint. This
change simply moves RegisterAggregator to the main sqlite3.go file,
and moves accompanying unit tests.

The rationale for this move is that it was not possible for downstream
using packages to use RegisterAggregator without also specifying (and
notifying the user) the 'trace' build tag.


kenshaw:
It's worth noting that (philosophically) this should probably be used/tested in the same place as RegisterFunc. Also, worth mentioning is that the additional GOTAGS=trace environment in the travis-ci configuration is no longer necessary, as the *_trace_test.go only tested the RegisterAggregator func.

kenshaw:
Just bumping this to determine if this will be merged or not. If there is anything that I need to change in this PR to get it included, please let me know. Thanks!

mattn:
LGTM. I'll check more something for merge.

mattn:
Thank you!

kenshaw:
Thanks for the merge -- really appreciate all the work you do @mattn on this project and others. Cheers!

PR #490, opened 2017-11-22
drop support of go1.6

mattn:


mattn:
already exists https://github.com/mattn/go-sqlite3/pull/462

PR #500, opened 2017-11-29
upstream 3.21.0 with UPDATE/DELETE LIMIT support

temoto:
Amalgamation built with
```
CFLAGS='-std=gnu99 -DSQLITE_ENABLE_RTREE -DSQLITE_THREADSAFE=1 -DSQLITE_ENABLE_FTS3 -DSQLITE_ENABLE_FTS3_PARENTHESIS -DSQLITE_ENABLE_FTS4_UNICODE61 -DSQLITE_TRACE_SIZE_LIMIT=15 -DSQLITE_DISABLE_INTRINSIC -Wno-deprecated-declarations' ./configure --enable-rtree --enable-fts3 --enable-fts4 --enable-json1 --enable-update-limit
```
maybe defining flags before configure was too much, sorry for mess.

Originally proposed in 
https://github.com/mattn/go-sqlite3/issues/213
https://github.com/mattn/go-sqlite3/pull/215


temoto:
Sorry, I merged and deleted branch in my repo. This patch is still relevant.

mattn:
You can use json1 with `go build -tag json1`

mattn:
please drop json1 feature from this PR

mattn:
please drop updating of sqlite3-binding.c and `#cgo CFLAGS: -DSQLITE_ENABLE_JSON1`.

temoto:
@mattn done

temoto:
Sorry I have script that deletes branches after merge.

Ping. Is everything OK?

temoto:
@mattn ping

mattn:
please drop changes of sqlite3-binding.c. I think this is not require changes for it.

temoto:
Amalgamation contains changes to query parser, adding `order by` and `limit` clause to `update, delete` statements. Without changes to sqlite3-binding.c it gives error:

```sql
update queue
  set locked = current_timestamp
  where (locked is null) and (scheduled < current_timestamp)
  order by scheduled limit ?

-- near "order": syntax error`
```

temoto:
https://www.sqlite.org/compile.html

> SQLITE_ENABLE_UPDATE_DELETE_LIMIT
> This option enables an optional ORDER BY and LIMIT clause on UPDATE and DELETE statements.
> If this option is defined, then it must also be defined when using the 'lemon' tool to generate a parse.c file. Because of this, this option may only be used when the library is built from source, not from the amalgamation or from the collection of pre-packaged C files provided for non-Unix like platforms on the website.

mattn:
Well, 3.21.0 is already merged in master branch. https://github.com/mattn/go-sqlite3/commit/c41df79ec0fe2cb5886b24dc9a203bf7021873bb

Where did you download from?

temoto:
I'm not saying there was version change, I just compiled amalgamation with update-limit flag.

`./configure --enable-rtree --enable-fts3 --enable-fts4 --enable-update-limit`

Code is taken from here http://www.sqlite.org/2017/sqlite-src-3210000.zip
Snapshop of the complete (raw) source tree for SQLite version 3.21.0. See How To Compile SQLite for usage details.
(sha1: 1892ebbac215095351a0ecc0c60a36e4d03104bc)

mattn:
sqlite3-binding.c have some `#ifdef SQLITE_ENABLE_UPDATE_DELETE_LIMIT`. So if it doesn't work, something should be wrong.

temoto:
Please pay attention to this quote by sqlite developers.

> If this option is defined, then it **must also be defined** when using the 'lemon' tool to generate a parse.c file. Because of this, **this option may only be used when the library is built from source**, not from the amalgamation or from the collection of pre-packaged C files provided for non-Unix like platforms on the website.

Amalgamation source they provide was configured without this flag. So it contains ifdefs to have update/delete order/limit feature to work, but it does not contain parser code to allow order/limit words in SQL statements. Because parser code is finished after `make sqlite3.c`.

Yes, it could be better for sqlite devs to add something like

```c
#ifdef SQLITE_ENABLE_UPDATE_DELETE_LIMIT
#error must configure with --enable-update-limit
#endif
```

I will try to send this patch, but it's only helpful for debugging the problem. We don't have a problem like "we don't know why it doesn't build". We know why it doesn't build, we know how to make it build and we know how to make it work.

temoto:
`sqlite-src-321000/README.txt`
> Other preprocessor defines
> --------------------------
> Additionally, preprocessor defines may be specified by using the OPTS macro
on the NMAKE command line.  However, not all possible preprocessor defines
may be specified in this manner as some require the amalgamation to be built
with them enabled (see http://www.sqlite.org/compile.html). For example, the
following will work:
 `  "OPTS=-DSQLITE_ENABLE_STAT4=1 -DSQLITE_ENABLE_JSON1=1"`
 However, the following will not compile unless the amalgamation was built
with it enabled:
 `  "OPTS=-DSQLITE_ENABLE_UPDATE_DELETE_LIMIT=1"`

mattn:
Thanks your pointing this. I understood now. However, please take a while for me. I hope that I'll change sqlite3-binding.c in my self.

mattn:
sqlite3 amalgamation code is large for reviewing changes. If, in the future, someone send me a pull-request of changes on sqlite3-binding.c, I may not be able to review it correctly. If someone(not you) is malicious, I might merge something. So I want that I'll change sqlite3-binding.c in my self. Sorry.

temoto:
I understand of course. Instead of reviewing changes you can just regenerate amalgamation from official sqlite source. I can patch your upgrade tool if you like.

GJRTimmer:
Should be closed because of #564

mattn:
Yes, but note that not mean that this enabling delete limit also when using  libsqlite3 on system always.

mattn:
gg

PR #536, opened 2018-02-24
README: Update list issues concurrent read/writes

ifraixedes:
Update the README FAQ section adding one issues more to the list of the
issues related with the concurrent reads and writes. This new added
issues gives a lot of insights, much more than the ones which where
already present.


mattn:
Thank you

PR #537, opened 2018-03-06
Add support for pre-update hooks

rbranson:
Enables pre-update calls in SQLite by changing the compile-time option. I couldn't find a documented downside to enabling it. Also pulls in some code from database/sql for type coercion when grabbing row data during the hook call.


mattn:
Why close?

PR #538, opened 2018-03-12
WIP: add basic sqlite command interpreter

anarcat:
While sqlite obviously ships its own commandline interface to operate
on databases, it is not always as widely available as one would
expect. For example, I'm working on an embeded device here that runs a
custom Linux distribution, and sqlite is not shipped with the
distro. Deploying a binary of sqlite fails because it depends on
shared libraries, and going down that rabbit hole is a problem that
can be pretty hairy to solve. And compiling sqlite itself statically
isn't quite workable for all sorts of reasons.

Which brings us to this project: golang compiles statically by
default, and I've successfully deployed a small program using this
library on the embedded device without problems. But to do some more
debugging, I now need an actual SQL shell to run commands.

So I wrote this small command that creates a go-sqlite3 binary which
behaves somewhat like the normal sqlite command. It's way more limited
of course: output isn't really formatted correctly into tables and
only crude functionality is implemented. But databases can be read
with SELECT queries, which can be passed on the commandline or through
a crude shell-like commandline prompt.

This is similiar to the commands in _example but is actually usable,
so I have thrown it in the standard `cmd` subdirectory, in the hope it
can be installed directly.

Still missing:

 - [ ] commandline history. can be implemented with [ishell] or
   [readline] or [GNU readline]
   
 - [ ] table formatting. possibly: [tabwriter], [tablewriter],
   [columnize], [cli-table]

[ishell]: https://github.com/abiosoft/ishell
[readline]: https://github.com/chzyer/readline
[GNU readline]: https://github.com/bobappleyard/readline
[tabwriter]: https://golang.org/pkg/text/tabwriter/
[tablewriter]: https://github.com/olekukonko/tablewriter
[columnize]: https://github.com/ryanuber/columnize
[cli-table]: https://github.com/geckoboard/cli-table

### DIFF
--- a/cmd/go-sqlite3/main.go
+++ b/cmd/go-sqlite3/main.go
@@ -1,0 +1,71 @@
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	_ "github.com/mattn/go-sqlite3"
+	"log"
+	"os"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		log.Fatal("not enough arguments")
+	}
+	path := os.Args[1]
+	query := os.Args[2]
+	db, err := sql.Open("sqlite3", path)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer db.Close()
+	runQuery(db, query)
+}
+
+func runQuery(db *sql.DB, query string) (err error) {
+	rows, err := db.Query(query)
+	if err != nil {
+		log.Println(err)
+		return err
+	}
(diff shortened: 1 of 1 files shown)


anarcat:
and just now, obviously, I find out about [usql](https://github.com/xo/usql) which makes this kind of moot... but i guess it could still be useful if people don't want the 16MB binary.. :)

feel free to close this at any time...

GJRTimmer:
Currently beyond the scope of this package.

PR #549, opened 2018-04-17
Fix linter issues

mjtrangoni:
Hi @mattn,

I fixed some `linter` issues. This could be maybe added to `Travis-CI` running [gometalinter](https://github.com/alecthomas/gometalinter) (of course disabling many checks).

If you are interested, I can submit a PR on this. Thanks!


sqlite3.go
```diff
@@ -1039,7 +1045,7 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 		case C.SQLITE_INTEGER:
 			val := int64(C.sqlite3_column_int64(rc.s.s, C.int(i)))
 			switch rc.decltype[i] {
-			case "timestamp", "datetime", "date":
+			case timestamp, datetime, date:
```
mattn: Is this correct?

sqlite3.go
```diff
@@ -1086,7 +1092,7 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 			s := C.GoStringN((*C.char)(unsafe.Pointer(C.sqlite3_column_text(rc.s.s, C.int(i)))), C.int(n))
 
 			switch rc.decltype[i] {
-			case "timestamp", "datetime", "date":
+			case timestamp, datetime, date:
```
mattn: ditto

sqlite3_test.go
```diff
@@ -1410,10 +1410,8 @@ func BenchmarkCustomFunctions(b *testing.B) {
 		sql.Register("sqlite3_BenchmarkCustomFunctions", &SQLiteDriver{
 			ConnectHook: func(conn *SQLiteConn) error {
 				// Impure function to force sqlite to reexecute it each time.
-				if err := conn.RegisterFunc("custom_add", customAdd, false); err != nil {
-					return err
-				}
-				return nil
+				err := conn.RegisterFunc("custom_add", customAdd, false)
```
mattn:
```go
return conn.RegisterFunc("custom_add", customAdd, false)
```

mjtrangoni: Good catch! I will be fixing it.

sqlite3.go
mjtrangoni: yes, I defined them as constant upstairs.

mjtrangoni: done! PTAL

mattn:
Thanks

PR #558, opened 2018-05-12
Adds GetFilename (sqlite3_db_filename) to SqliteConn

zgramana:
Enables a user with a `SqliteConn` reference to find out, among other things, what it's a connection to without having to manually track connection handles. Also very handy for diagnostic purposes.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1018,6 +1018,17 @@ const (
 	SQLITE_LIMIT_WORKER_THREADS      = C.SQLITE_LIMIT_WORKER_THREADS
 )
 
+// GetFilename returns the absolute path to the file containing
+// the requested schema. When passed an empty string, it will
+// instead use the database's default schema: "main".
+// See: sqlite3_db_filename, https://www.sqlite.org/c3ref/db_filename.html
+func (c *SQLiteConn) GetFilename(schemaName string) string {
+	if schemaName == "" {
+		schemaName = "main"
+	}
+	return C.GoString(C.sqlite3_db_filename(c.db, C.CString(schemaName)))
+}
+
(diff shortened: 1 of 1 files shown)


sqlite3.go
```diff
@@ -1018,6 +1018,17 @@ const (
 	SQLITE_LIMIT_WORKER_THREADS      = C.SQLITE_LIMIT_WORKER_THREADS
 )
 
+// GetFilename returns the absolute path to the file containing
+// the requested schema. When passed an empty string, it will
+// instead use the database's default schema: "main".
+// See: sqlite3_db_filename, https://www.sqlite.org/c3ref/db_filename.html
+func (c *SQLiteConn) GetFilename(schemaName string) string {
+	if schemaName == "" {
```
otoolep: Any particular reason you decided to actually do this? Why not just let an empty string be passed to the underlying function?

zgramana: Yes. Passing an empty string results in a error. Most `sqlite` users are unaware of a sqlite database connection to have multiple named schemas open (since few use attached tables), and consequently don't know that the default is `main`.

Most people will just get confused, so handling the empty string allows a more intuitive, less surprising, user experience.

zgramana: That said, I also considered splitting into two `func`s, something like:

- `GetFilename()` which is a convenience function for
- `GetFilenameFor(schema string)`

GJRTimmer:
@zgramana Could you please rebase, and add a test based which uses the `TempFilename(..)` and check if the returned filename is the same. `sqlite3_db_filename` should return an absolute path.

zgramana:
Yes, I can do that.

PR #564, opened 2018-05-23
Fix #213

### LINKED ISSUE #213: Is there any way to use the driver with SQLITE_ENABLE_UPDATE_DELETE_LIMIT flag?

GJRTimmer:
ADD: SQLITE_ENABLE_UPDATE_DELETE_LIMIT

Fixes: #213


mattn:
Thank you

EmpireJones:
I'm trying to interpret this comment:
https://github.com/mattn/go-sqlite3/pull/215#issuecomment-391655257
> note that not mean that this enabling delete limit also when using libsqlite3 on system always.

Am I understanding correctly that this feature is **only** available with usage of a separate build of libsqlite3, and **not** the "linux" tag build, as that one uses the 'amalgamation'?

mholt:
How does one use this change? I get `near "LIMIT": syntax error` on the DELETE statement used in the test case.

rittneje:
https://www.sqlite.org/compile.html#enable_update_delete_limit

"If this option is defined, then it must also be defined when using the Lemon parser generator tool to generate a parse.c file. Because of this, this option may only be used when the library is built from source, not from the amalgamation or from the collection of pre-packaged C files provided for non-Unix like platforms on the website."

Just adding the compile-time option is not sufficient. You would need a custom amalgamation.

mholt:
I see, thank you for the tip!

PR #565, opened 2018-05-23
ADD: SQLITE_DEFAULT_WAL_SYNCHRONOUS

GJRTimmer:
FROM: [Sqlite3.org](https://www.sqlite.org/compile.html)

SQLITE_DEFAULT_WAL_SYNCHRONOUS=1. For maximum database safety following a power lose, the setting of PRAGMA synchronous=FULL is recommended. 

However, in WAL mode, complete database integrity is guaranteed with PRAGMA synchronous=NORMAL. With PRAGMA synchronous=NORMAL in WAL mode, recent changes to the database might be rolled back by a power loss, but the database will not be corrupted. 

Furthermore, transaction commit is much faster in WAL mode using synchronous=NORMAL than with the default synchronous=FULL. For these reasons, it is recommended that the synchronous setting be changed from FULL to NORMAL when switching to WAL mode. 

***This compile-time option will accomplish that.***

Summary: When a user uses this library and sets the PRAGMA journal_mode to `wal` then this option should be default be set.


mattn:
I'm not well known about this flag. Do this work without performances less?

GJRTimmer:
Not a problem I will explain.
Actually it increases performance.

# What does this FLAG do
This FLAG has no effect on any of the journaling modes except for WAL.

Normally for all the journaling modes except WAL you must set `PRAGMA synchronous=FULL` to garantuee not losing any data. (This is the default setting by the way)

When you configure `WAL` mode you configure this with:
`PRAGMA journal_mode=wal` however a lot of people do not know that to correctly configure `WAL` mode you also have to set `PRAGMA synchronous=NORMAL` because having `WAL` mode with synchronous=FULL is redundant and actually descreases performance.

This FLAG will set synchronous to `NORMAL` when a user activates the journal `WAL` mode. So it will automatically set the synchronous mode of SQLite to `NORMAL`. This is only for the `WAL` mode. 

See this as a conveniance FLAG that will set the synchronous mode to `NORMAL` for users who want to use the `WAL` mode and who do not know that for using `WAL` you also have to set `PRAGMA synchronous=NORMAL` to get the full performance of the `WAL` mode.

# Summary
This is safe to activate and will actually increase performance for the journal `WAL` mode and has no effect or changes for any of the other journaling modes. Also when using `WAL` mode with synchronous mode `NORMAL` you will not loose any of your data.

@mattn I hope this helps you understand; its completly safe and increases performance.

giner:
The benchmarks above seem to be irrelevant here as they don't cover writes.

PR #566, opened 2018-05-23
ADD: SQLITE_OMIT_DEPRECATED

GJRTimmer:
FROM: [Sqlite3.org](https://www.sqlite.org/compile.html)

Omitting deprecated interfaces and features will not help SQLite to run any faster. It will reduce the library footprint, however.


mattn:
This is good to make binary smaller and check strictly deprecated APIs. Thanks.

PR #572, opened 2018-05-27
Update/docs

GJRTimmer:
* Exclude .vscode settings from repository
* Updated TOC, TOC did not match README, Moved Alpine section to Linux Subsection to match README.


mattn:
Thank you

Issue #542, opened 2018-03-24
nil slice to zero-length blob


lucasmrod:
Changes to fix https://github.com/mattn/go-sqlite3/issues/424 also make `nil` slice be converted to zero-length blob. Instead, `nil` slice should be converted to value `NULL`.
```
Current code:
	[]byte -> zero length blob
	nil    -> zero length blob
Proposal:
	[]byte -> zero length blob
	nil    -> NULL blob
```
I can submit a PR if the proposal makes sense to you.

GJRTimmer:
@lucasmrod makes sense, can you submit a PR of evaluation and testing, please also include tests in your PR.

PR #583, opened 2018-05-31
Add nil check in bind and a test

lucasmrod:
This is the fix for #542. 

There are extensive tests in the existing `TestNilAndEmptyBytes`, but we cannot test the change in there because of how `sqlite3_column_blob()` works: _"The return value from sqlite3_column_blob() for a zero-length BLOB is a NULL pointer."_ -> this means that both `NULL` and `zero-length blob` map to a `nil` byte slice. So... I ended up adding a separate test `TestInsertNilByteSlice` to check that storing a nil byte slice is now storing a `NULL`.

### DIFF
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -1593,6 +1593,10 @@ func TestInsertNilByteSlice(t *testing.T) {
 	if _, err := db.Exec("insert into blob_not_null (b) values (?)", nilSlice); err == nil {
 		t.Fatal("didn't expect INSERT to 'not null' column with a nil []byte slice to work")
 	}
+	zeroLenSlice := []byte{}
+	if _, err := db.Exec("insert into blob_not_null (b) values (?)", zeroLenSlice); err != nil {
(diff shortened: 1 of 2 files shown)


sqlite3_test.go
```diff
@@ -1580,6 +1580,21 @@ func TestNilAndEmptyBytes(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	defer db.Close()
+	if _, err := db.Exec("create table blob_not_null (b blob not null)"); err != nil {
+		t.Fatal(err)
+	}
+	var nilSlice []byte
+	if _, err := db.Exec("insert into blob_not_null (b) values (?)", nilSlice); err == nil {
+		t.Fatal("didn't expect INSERT to 'not null' column with a nil []byte slice to work")
+	}
```
mattn: Could you please add test inserting nil too?

mattn:
Thank you!

GJRTimmer:
@lucasmrod Thanks, great work

PR #589, opened 2018-06-09
bump sqlite3 3.24.0 and add TestUpsert.

tuxlinuxien:


GJRTimmer:
@tuxlinuxien will try and test it today. Any other new features we should be looking into ?

tuxlinuxien:
@GJRTimmer the only thing I am aware of is the [UPSERT feature](https://www.sqlite.org/draft/releaselog/3_24_0.html). However is page is still marked as _draft_ so maybe more official features will be announced later.

GJRTimmer:
@tuxlinuxien Because this in an update of sqlite we probably also should look into the new Auxiliary columns. Do you think we need to update the virtual table and testing to support this out-of-the-box ?

tuxlinuxien:
> Do you think we need to update the virtual table and testing to support this out-of-the-box ?

@GJRTimmer to be honest my knowledge are limited about the core of sqlite3 so I am not the right person to answer this question. Sorry

GJRTimmer:
@tuxlinuxien Not a problem, I think UPSERT is the most important one, which people are waiting for.
The auxiliary columns feature can be implemented later on.

I've requested one change to your code, which is solid and very nice. Change I've requested is to limit the output lines of the test when its skipped.

tuxlinuxien:
@GJRTimmer done. Just need to wait for travis to check this up.

```sh
$>go test --tags='libsqlite3' -v ./. # local libsqlite3 3.22.0
[...]
=== RUN   TestUpsert
--- SKIP: TestUpsert (0.00s)
	sqlite3_test.go:237: UPSERT requires sqlite3 => 3.24.0
[...]
```

GJRTimmer:
@tuxlinuxien Will merge after Travis-CI

PR #591, opened 2018-06-12
fix/118

### LINKED ISSUE #118: Dead code in SQLiteRows.Next
Hello,
The following test (line 502 of sqlite.go) is useless:

```
                    switch dest[i].(type) {
                    case sql.RawBytes:
                            dest[i] = (*[1 << 30]byte)(unsafe.Pointer(p))[0:n]
                    default:
```

Just take a look at the caller:
http://golang.org/src/pkg/database/sql/sql.go?s=38519:38546#L1492

```
if rs.lastcols == nil {
    rs.lastcols = make([]driver.Value, len(rs.rowsi.Columns()))
}
rs.lasterr = rs.rowsi.Next(rs.lastcols)
```

The final/target type is not known when driver.Rows.Next is called...

Regards.

GJRTimmer:
* Added TestMultiBlobs
* Removed dead code

Fixes #118

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1937,8 +1937,6 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 			}
 			n := int(C.sqlite3_column_bytes(rc.s.s, C.int(i)))
 			switch dest[i].(type) {
-			case sql.RawBytes:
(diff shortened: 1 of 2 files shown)


mattn:
LGTM

mattn:
Thank you

PR #596, opened 2018-06-12
Update Travis-CI

GJRTimmer:
Moved `sqlite_vacuum_incr` and `sqlite_vtable` to module job.

* `sqlite_vacuum_incr` will only store the required auto vacuum information into the database page, and not preform an vacuum therefor this can be added easily to the module job.
* `sqlite_vtable` does not require an separate job because user will include the tag in combination with others.

This change will approximately save 24+ minutes on Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,8 @@ env:
   matrix:
     - GOTAGS=
     - GOTAGS=libsqlite3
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

PR #597, opened 2018-06-13
Add support for pre-update hooks

rbranson:
This PR adds support for pre-update hooks. Pre-update hooks allow getting the before & after row images for a given mutation. The standard update hook is insufficient for many use cases. Some careful orchestration would allow getting the new row image by ID in the insert & update cases, in the delete case, the data is already gone.

The only caveat here is that it does require compiling SQLite with the SQLITE_ENABLE_PREUPDATE_HOOK directive. Given that it is not the default, one would think there is additional risk or a performance penalty involved.

Issue #598, opened 2018-06-15
2.0.0


gjrtimmer:
# Go-SQLite3 v2.0.0 

***WORK IN PROGRESS***

# Please Submit Your Wishes / Requests

## Planned Updates
* [ ] Add Context to errors [More Information](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully)
* [x] Add additional tests for provided options
* [x] Fix Extension loading; current when a module cannot be found it returns an unknown error 115 from SQLite; provided more error information which extension could not be loaded.
* [x] Implement Golang:1.10 SQL Connector
* [x] Rewrite connection configuration
* [x] See if sqlite3.go can be split on a functional level with moving to code for statements to its own file
* [x] Update Examples
* [x] Research if in combination with new configuration if the DSN connection can be configured with options without a `_` prefix.
* [x] Add more comments to explain workings to allow easier contribution
* [x] Add contributing guide
* [x] Increase code coverage
* [x] Implement Golang project layout

## Tests
Write additional tests

* [x] Foreign Keys
* [ ] FTS5
* [ ] ICU
* [x] Introspect
* [ ] JSON1
* [ ] Secure Delete (If possible)
* [X] Stat4
* [x] User Auth

GJRTimmer:
@mattn Please checkout and take a look, documentation is not done yet. Can you provide some feedback for me ?

mattn:
Thanks your great works. I'm thinking all of task are not required for 2.0.0 .

GJRTimmer:
@mattn Will update the tasks

GJRTimmer:
@mattn I will complete the 2.0.0 and incorporate all the current missing items like documentation.

GJRTimmer:
@mattn I would like to know how do you feel about introducing a dependency on `github.com/pkg/errors`  so we can wrap errors and provide more and better error information.

mattn:
I don't have strong opinion but what the use-case?

y4dkit:
Is there a support for Update Hook? (https://www.sqlite.org/c3ref/update_hook.html)

Is it possible to set a callback whenever I insert, update or delete?

GJRTimmer:
@mattn Use case is to better track down errors; we can add an additional context error where it occurs. So when printing the stacktrace you can better track down where it originated.

GJRTimmer:
@y4dkit Will look into it; I will try so see if I at least can inject the open Pre-Update Hook.
[...]

PR #487, opened 2017-11-21
[WIP] add support SEE (crypto key)

mattn:

PR #611, opened 2018-07-20
v2.0.0

GJRTimmer:
Update driver to v2.0.0

* Complete Rewrite
* Move all documentation to Wiki
* Extendend documentation
* Repaired examples
* Provided README for all examples including build commands
* Implemented Golang Project layout
* Implemented Pre-Update Hook from #597 
* Written Additional Tests
* Increased coverage
* Implemented missing functionality for Go:1.10.x
  * *Config
  * `driver.Connector`
  * `driver.DriverContext`
* Rewrite of DNS Options
* Rewrite of CryptEncoder
  * User can now easily create own `CryptEncoder` and register is with the driver so it can be used from the DSN.
* Added support for PPC Architecture
* Rewrite of the BackupTests because o Go:1.8
* Add tests for several features
* Fix race error
* Added GitHub Templates

This will close PR #597 and Issue #598


GJRTimmer:
This requires manual merge because the master has to be overwritten with this new branch.

GJRTimmer:
PR #439 and PR #487 will be done in the new v2.0.0; I will do those by hand and close the corresponding PR.

mattn:
I prefer to use error message leading with lower case letters. For example: `failed to ...` not `Failed to ...`. If I had wrote upper case, please fix in this PR.

GJRTimmer:
I understand not a problem. I will fix this and add it to the contributing guide. Anything else you want or prefer? I honestly don't mind because you started this great repo. Any rules or requirements you want shall be added to the contributing guide which I have added in this PR.

mattn:
As far as I can see, patckage path is moved to `driver` directory, So this break compatibility?

GJRTimmer:
Yes, and second what breaks backwards compatibility are the names of the DSN. No longer is prefix '_' required for the options within the DSN, because there are no conflicts.

Check out the Wiki.

GJRTimmer:
@mattn Also several options for the DSN have been renamed. `_auth_user` for example is now simple `user`. On the Wiki page of the DSN I have even created a conversion list between pre-2.0.0 versions and the 2.0.0 version.

See [Conversion List](https://github.com/mattn/go-sqlite3/wiki/DSN#conversion-table)

GJRTimmer:
Also `_loc` is now fully implemented as `tz` (Timezone). Within the DSN as well within the Config.

mattn:
I'm feeling that the impact of this breaking backward compatibility for third party packages is too big. Most of app or packages using go-sqlite3 is not managed by package manager like glide, dep or vgo. I wonder this proposal requires somewhat unreasonable work for users. May be, we will get many vote down comments. :-(

So, how about to separate repository? Move to github.com/go-sqlite3/v1 or v2, Or use pkg.in ?

BTW, your works is great. Thanks.

GJRTimmer:
@mattn Thanks. We could do a separate repository however I'm not a fan for it.
I do not think that we will receive a lot of votes down. 

We have significantly increased the documentation. Provided conversion tables and we follow Semantic Versioning. We because things are not entirely backwards compatible we increased the Major version to `2`.

Users can also always checkout tag 1.9.0; I honestly think users will be happy with a major update. I think we should update it and if we really get much backfire; we can either look into a rollback or providing the community with additional conversion documentation.

GJRTimmer:
BTW, Native full encryption (SEE), build upon the v.2.0.0 is almost done

GJRTimmer:
@mattn see previous post; If we really get to many complaints we always can create the separate repository  like to suggest after the complaints; I really think we should update it and just see.

Adding a large text to the beginning to the README as a warning about the backwards compatibility should help to.

GJRTimmer:
@mattn Any thought yet.

P.S. I've completed SEE native Golang encryption by incorperating any other repo and makes quite some change to get it all working. But it is working perfectly. Also I have several other updates ready.

mattn:
Could you please explain all of differences between v1 and v2 ? I want to know how many breaking backword compatibilities. Thanks.

GJRTimmer:
@mattn 

* `_` Prefix removed from DSN configuration keys
* Renamed DSN config key `loc` to `tz` and added `timezone` as alternative
* Dropped config key `_auth`
* Removed prefix `_auth_` from UserAuthentication config keys
* Implemented Golang project structure; moved the driver to folder driver, this changes the import path. package name `sqlite3` remains the same. Import path is now `github.com/mattn/go-sqlite3/driver`

| Pre-2.0.0 | 2.x |
|-----------|-----|
| cache | cache |
| mutex | mutex |
| _auth | N.A. |
| _auth_user | user |
| _auth_pass | pass |
| _auth_salt | salt |
| _auth_crypt | crypt |
| _auto_vacuum | auto_vacuum |
| _busy_timeout | busy_timeout |
| _case_sensitive_like | case_sensitive_like |
| _cslike | cslike |
| _defer_foreign_keys | defer_foreign_keys |
| _defer_fk | defer_fk |
| _foreign_keys | foreign_keys |
| _fk | fk |
| _ignore_check_constraints | ignore_check_constraints |
| _journal_mode | journal_mode |
| _journal | journal |
| _loc | tz, timezone |
| _locking_mode | locking_mode |
| _locking | locking |
| _query_only | query_only |
| _recursive_triggers | recursive_triggers |
| _rt | rt |
| _secure_delete | secure_delete |
| _synchronous | synchronous |
| _sync | sync |
| _timeout | timeout |
| _txlock | txlock |
| _vacuum | vacuum |
| _writable_schema | writable_schema |

Additional documentation moved from README to Wiki Pages.

mattn:
> _ Prefix removed from DSN configuration keys

I'm thinking this might occur problem when sqlite include new flag which have same name in go-sqlite3.

> Implemented Golang project structure; moved the driver to folder driver, this changes the import path. package name sqlite3 remains the same. Import path is now github.com/mattn/go-sqlite3/driver

Sorry, I still understand well why this is needed.

GJRTimmer:
>> _ Prefix removed from DSN configuration keys

>I'm thinking this might occur problem when sqlite include new flag which have same name in go-sqlite3.

The chance that a conflict will arise will be small because the keys that are given to SQLite are for low-level configuration. Currently there is no such change on the roadmap. Second if their might be a conflict in the future we can easily change it. I think however that the chance will be extremely slim that we will have a conflict in the future.

>> Implemented Golang project structure; moved the driver to folder driver, this changes the import path. package name sqlite3 remains the same. Import path is now github.com/mattn/go-sqlite3/driver

>Sorry, I still not understand well why this is needed.

This was done by my choice, I you want to revert it; it will be fine. To explain my decision. This layout will give us the opportunity to add additional folders and `applications`, `libraries` to the project.

What I was thinking about is to add a Web UI service / Shell to the project in the future. In which you could manage your SQLite database. Available as shell program and library so you can embed it within you project. If you look at the README of `https://github.com/golang-standards/project-layout` then you will see where I got the idea.

If done it with my eye on the future to expand en develop this repository even further. If you do not want this. It will be fine with me.

GJRTimmer:
@mattn  See privous post for explanation.

@mattn If you want to see the end result. I've updated the master of my `own` fork to v2.0.0 including the latest fixes for the features from `fix/features` as well as the new `SEE` full database encryption.

You can find the documentation for it within my Wiki Pages. I hope you like it. [Take a look](https://github.com/GJRTimmer/go-sqlite3)

GJRTimmer:
@mattn within my own repo I've reversed the SEE; this requires some additional work.

mattn:
> What I was thinking about is to add a Web UI service / Shell to the project in the future. In which you could manage your SQLite database. Available as shell program and library so you can embed it within you project. If you look at the README of Golang Project Layout then you will see where I got the idea.

I prefer to add `cmd/sqlite-shell` or something. I am concerned about changing the import path.

Since I am conservative, I always think that I would like to prevent existing users from asking for changes.

Please give me more time to think about this.

GJRTimmer:
@mattn Not a problem take all the time you need. Please check out other SQL drivers like; Postgres, MySQL maybe those will help you choose what to do.

PR #616, opened 2018-08-12
Update README: URL-encode connection options

jung-kurt:
This change could save users from having to dig into source code to see how connection options are parsed.


mattn:
Thank you

PR #633, opened 2018-09-16
Mention correct algorithm in function docs.

israel-lugo:
CryptEncoder SHA384 and SHA512 variants (with and without salt) were incorrectly documented as using SHA256.


mattn:
Thank you

PR #641, opened 2018-09-22
Actually check the value of RunTests

### LINKED ISSUE #640: Return value of testing.RunTests is not checked
Therefore, if a test fails with a non-fatal error, TestSuite will still appear to pass.

PR incoming...

akalin:
This fixes #640.


akalin:
Done! Please take another look.

mattn:
Thank you.

PR #643, opened 2018-09-22
Distinguish between NULL and zero-length blobs on query

### LINKED ISSUE #642: NULL vs. zero-length blobs are not distinguished on query
#542 fixed the insert case. I'll file a PR to fix the query case.

akalin:
Map NULL to nil and non-NULL zero-length blobs to a zero-length byte slice.

This fixes #642.


akalin:
Note that this test failing (and some others) won't show up until #640 is fixed.

mattn:
@GJRTimmer How do you think this?

mattn:
Thanks

Issue #238, opened 2015-09-16
C.struct_Mem larger than address space


SuperGod:
I get this error information  when build go-sqlite3 on windows use go of i386,both go1.4 and go 1.5

```
.\callback.go:31: type [1073741824]*C.struct_Mem larger than address space
.\callback.go:31: type [1073741824]*C.struct_Mem too large
.\callback.go:38: type [1073741824]*C.struct_Mem larger than address space
.\callback.go:38: type [1073741824]*C.struct_Mem too large
```

mattn:
could you please show me the coe?

SuperGod:
I just run this command：
```
go get -u github.com/mattn/go-sqlite3
```
and then get this error information.

maybe this code is error?

```
func callbackTrampoline(ctx *C.sqlite3_context, argc int, argv **C.sqlite3_value) {
	args := (*[1 << 30]*C.sqlite3_value)(unsafe.Pointer(argv))[:argc:argc]
	fi := (*functionInfo)(unsafe.Pointer(C.sqlite3_user_data(ctx)))
	fi.Call(ctx, args)
}
```
1 << 30 is 1073741824.
when I change "1<<30" to "1<<20",then no errors.
so I think maybe "1 << 30 " is to large in 32bit platform

mattn:
Ah, overflow. I'll fix.

mattn:
@SuperGod could you please try below?

```go
package main

func main() {
	if false {
		a := 1 << 30
		println(a)
	}
}
```

SuperGod:
nothing  print, My go version:

```
go version go1.5.1 windows/386
```

mattn:
callbackTrampoline is not called with `go get`, i think.

SuperGod:
Maybe this example will help you:
```
package main

import (
	"fmt"
)

func main(){
	var array [1<<30]int
	fmt.Println(len(array))
}
```

build this code,I get this errors:
```
.\array.go:8: type [1073741824]int larger than address space
.\array.go:8: type [1073741824]int too large
```

The type of "int" take 4 Byte memory,so the max slice/array length will be MaxUnit/sizeof(int) or MaxInt/sizeof(int)

SuperGod:
this code build with no errors.
```
package main

import (
	"fmt"
	"math"
)

func main(){
	var array [math.MaxInt32-1]byte
	fmt.Println(len(array))
}
```

but this code will cause error:
```
package main

import (
	"fmt"
	"math"
)
[...]

PR #654, opened 2018-10-23
Fix build failure on Windows/386

typeless:
The array size exceeds the address space of 32-bit platform.
See https://github.com/mattn/go-sqlite3/issues/238


typeless:
Even though this looks trivial, I don't have an environment with Windows/386 GCC compiler to test it. Maybe we should add Windows/386 to the CI pipeline?

mattn:
For now, I'll merge this.

PR #655, opened 2018-10-24
Fix build failure of incorrect expression for unsafe.Sizeof

typeless:
Emergency fix to make the CI happy again.
Tested on Linux/amd64.

I am still investigating how to test on Windows/386


mattn:
Thanks.

PR #657, opened 2018-11-02
travis.yml: build Go tip

kevinburke:
I got a compile failure building this project with Go tip, and
I figured it might be good to build continually, so we can ensure that
any errors are resolved before they get released to a wider audience.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,11 @@ go:
   - 1.9.x
   - 1.10.x
   - 1.11.x
+  - master
 
 before_install:
   - |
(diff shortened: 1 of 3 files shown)


kevinburke:
I expect this to fail due to #658, but at least that failure will fail the build.

kevinburke:
Yep, here's the reproduction: https://travis-ci.org/mattn/go-sqlite3/jobs/449672331

kevinburke:
Ah, I think the fix might actually be kind of easy, just amended the commit and testing it out now.

mattn:
Thanks!

PR #664, opened 2018-11-17
updating Windows TDM-GCC link to properly render

bk521234:
fixing "[here](ttps://sourceforge.net/projects/tdm-gcc/)" to "[here](https://sourceforge.net/projects/tdm-gcc/)".


mattn:
Thank you

bk521234:
thank YOU! go-sqlite3 is amazing.

PR #665, opened 2018-11-21
Fix gometalinter issues

mjtrangoni:
Hi @mattn, 

I fixed here some trivial errors reported by gometalinter.

### DIFF
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -1299,10 +1299,7 @@ func TestAggregatorRegistration(t *testing.T) {
 
 	sql.Register("sqlite3_AggregatorRegistration", &SQLiteDriver{
 		ConnectHook: func(conn *SQLiteConn) error {
-			if err := conn.RegisterAggregator("customSum", customSum, true); err != nil {
-				return err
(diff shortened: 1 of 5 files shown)


sqlite3_test.go
```diff
@@ -1299,10 +1299,7 @@ func TestAggregatorRegistration(t *testing.T) {
 
 	sql.Register("sqlite3_AggregatorRegistration", &SQLiteDriver{
 		ConnectHook: func(conn *SQLiteConn) error {
-			if err := conn.RegisterAggregator("customSum", customSum, true); err != nil {
-				return err
-			}
-			return nil
+			return conn.RegisterAggregator("customSum", customSum, true)
```
mattn: What warning on gometalinter?

mjtrangoni: Oops, here is the warning message,
```
sqlite3_test.go:1302:4:warning: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013) (megacheck)
````

mattn:
Thank you

Issue #667, opened 2018-12-05
Opening with rw mode doesn't work correctly


gonapps:
This code should return an error if there's no database file, but it doesn't

```go
    _, err := sql.Open("sqlite3", "file:database.db?mode=rw&_foreign_keys=true")
    if err != nil {
        fmt.Println(err)
    }
```

> SQLITE_OPEN_READWRITE
The database is opened for reading and writing if possible, or reading only if the file is write protected by the operating system. In either case the database must already exist, otherwise an error is returned.

Maybe it happens because of this line of the code?
[https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L1341](https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L1341)

mattn:
So you expect to remove SQLITE_OPEN_CREATE from flags?

gonapps:
Yes I do, when the flag is "rw"

rittneje:
Are you sure this is a bug? `sql.Open` itself doesn't necessarily call `Open` on the underlying driver. Try calling `Ping()` on the resulting database and see what it returns. The SQLite documentation itself says this should work as you expect.

https://www.sqlite.org/c3ref/open.html

>mode: The mode parameter may be set to either "ro", "rw", "rwc", or "memory". Attempting to set it to any other value is an error. If "ro" is specified, then the database is opened for read-only access, just as if the SQLITE_OPEN_READONLY flag had been set in the third argument to sqlite3_open_v2(). If the mode option is set to "rw", then the database is opened for read-write (but not create) access, as if SQLITE_OPEN_READWRITE (but not SQLITE_OPEN_CREATE) had been set. Value "rwc" is equivalent to setting both SQLITE_OPEN_READWRITE and SQLITE_OPEN_CREATE. If the mode option is set to "memory" then a pure in-memory database that never reads or writes from disk is used. It is an error to specify a value for the mode parameter that is less restrictive than that specified by the flags passed in the third parameter to sqlite3_open_v2().

javierbauza:
Maybe it's okay that it doesn't return an error, but I see that when mode=rw is passed, the database is still created if the file doesn't exist. Also, if an empty string is passed as the file name, an *sql.DB object is returned and the Ping method to this database does not return any errors.

PR #669, opened 2018-12-07
Fix RW mode in Open

gonapps:
Hi, Here is a pull request about this [issue](https://github.com/mattn/go-sqlite3/issues/667)


mattn:
please gofmt

PR #680, opened 2019-01-05
improve FAQ re: in-memory databases

rittneje:
Adding some additional useful information to the FAQ about how to properly use an in-memory database.


mattn:
Thank you

Issue #606, opened 2018-07-11
TEXT type read in as []byte


srahul123:
Following is my golang code:
package main
```
import "fmt"
import "database/sql"
import _ "github.com/mattn/go-sqlite3"

func main() {
        db, _ := sql.Open("sqlite3", "temp.db")

        db.Exec("CREATE TABLE X (col)")

        db.Exec("INSERT INTO X (col) VALUES (CAST('abc' AS TEXT))")

        rows, _ := db.Query("SELECT col FROM X")
        rows.Next()
        var col interface{}
        rows.Scan(&col)
        rows.Close()
        fmt.Printf("%T\n", col)
}
```
It prints _[]uint8_. Shouldn't it print _string_ ? _sqlite3 temp.db '.dump X'_ prints _INSERT INTO X VALUES('abc');_ . So, I think it is stored as TEXT in the DB

rittneje:
https://github.com/mattn/go-sqlite3/blob/3fa1c550ffa69b74dc4bfd5109b4e218f32c87cf/sqlite3.go#L2065
For some reason it casts the string into a byte slice before returning it. @mattn Any idea why it does this?

mattn:
Fixed: scanning a TEXT column into interface{} now yields a Go string (the old []byte cast is gone). Verified on current code. Closing as resolved.

PR #688, opened 2019-01-26
sqlite3_column_type SQLITE_TEXT returns string not []byte

dimroc:
Return as `string` as opposed to `[]byte` or `[]uint8` arrays.

This brings sqlite closer in line with other dbs like postgres, allowing downstream consumers to assume the scanned value is string across underlying dbs.

This implementation is a first draft. If it's good as is, great, if not, let's start a dialog on the issue and figure out how best to resolve it.

This change is related to https://github.com/jinzhu/gorm/issues/2276 and https://github.com/mattn/go-sqlite3/issues/606.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -223,7 +223,6 @@ var SQLiteTimestampFormats = []string{
 const (
 	columnDate      string = "date"
 	columnDatetime  string = "datetime"
-	columnText      string = "text"
 	columnTimestamp string = "timestamp"
 )
 
@@ -2062,14 +2061,8 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 					t = t.In(rc.s.c.loc)
 				}
 				dest[i] = t
-			case columnText:
-				dest[i] = s
 			default:
-				if strings.Contains(strings.ToLower(rc.decltype[i]), "char") {
-					dest[i] = s
-				} else {
-					dest[i] = []byte(s)
-				}
+				dest[i] = s
 			}
 		}
 	}
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -223,6 +223,7 @@ var SQLiteTimestampFormats = []string{
 const (
 	columnDate      string = "date"
 	columnDatetime  string = "datetime"
+	columnText      string = "text"
 	columnTimestamp string = "timestamp"
 )
 
@@ -2061,10 +2062,15 @@ func (rc *SQLiteRows) Next(dest []driver.Value) error {
 					t = t.In(rc.s.c.loc)
 				}
 				dest[i] = t
+			case columnText:
+				dest[i] = s
 			default:
-				dest[i] = []byte(s)
+				if strings.Contains(strings.ToLower(rc.decltype[i]), "char") {
+					dest[i] = s
+				} else {
+					dest[i] = []byte(s)
+				}
 			}
-
 		}
 	}
 	return nil


rittneje:
I really don't think we should be using the declared type of the column. That only works if you are just querying out a column directly. If you've got an expression (like appending two strings together, for example), the declared type comes back as the empty string, and you'll still have a problem. https://www.sqlite.org/capi3ref.html#sqlite3_column_decltype

I'm assuming the declared type logic was added to handle timestamps. (Unfortunately, timestamps aren't really their own type in SQLite, which is where the problem comes from.) But beyond that, I think we should just say that a string is a string.

rittneje:
The "declared type" logic is already in master. So I'm assuming it originally got added to this library to handle timestamps. @mattn is that a correct assessment?

SQLite itself doesn't return a string as `[]uint8`. That's this library's doing. SQLite attaches types to individual values, not columns. So by "a string is a string", I mean that if SQLite says the type of the value as reported by `sqlite3_column_type` is `SQLITE_TEXT`, this library should return the value as a `string`, not as a `[]byte`.

dimroc:
If I'm reading this code correctly @rittneje, that doesn't seem to be true:
https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L2039-L2066

Notice that the link above and my changes are inside the `case C.SQLITE_TEXT:` statement, which I believe you are referring to. #606 also brings this up.

rittneje:
I think there may be some confusion over what SQLite is. SQLite is a C library maintained by a totally separate team of people. https://www.sqlite.org/index.html

This library is a wrapper around the actual SQLite library that meets the database/sql/driver interfaces using cgo.

In short, what I'm saying is the `default` case in the code you linked should just be `return s`.

dimroc:
Ok great, I'm all for that. That solves my issue by returning a `string`, since that's what `s` is. I can update this PR to do just that.

dimroc:
@rittneje I am, however, concerned about this breaking change affecting those relying on the existing `[]byte` rather than `string` in their downstream `Scan` method. Is there a process in place in this repo to accommodate that, such as versioning?

rittneje:
It sounds like there are plans for a version 2. I don't know much about it though. https://github.com/mattn/go-sqlite3/issues/598

Note that people that were explicitly scanning into `[]byte` or `string` won't notice a difference. Only people that were scanning into `interface{}` would be affected. I'm not sure how common that really is outside these ORMs. https://golang.org/src/database/sql/convert.go?s=5759:5806#L199

My opinion is that this is really a bugfix, but I'd be curious what others have to say.

se3000:
What is the status of this? Would be nice to get it pulled in. Agree with @dimroc and @rittneje.

mattn:
Nothing will be change since the column type in query may be concating or calculating with another type. What query did you run? In SQLite, you have to make the type strictly. If you want to change default behavior, please add struct field in SQLiteDriver to make opt-in. And modify the flag in ConnectHook. A way to hook connection:

https://github.com/mattn/go-sqlite3/blob/3fa1c550ffa69b74dc4bfd5109b4e218f32c87cf/_example/hook/hook.go#L16

rittneje:
@mattn Here is a simple example that demonstrates the problem and the fix.
```go
var x interface{}
if err := db.QueryRow("SELECT 'hello'").Scan(&x); err != nil {
	panic(err)
}
fmt.Printf("%T", x)
```

With the current version of this library, it outputs `[]uint8` (i.e., `[]byte`). With the fix, it outputs `string`. This is because `sqlite3_column_type` gives back the actual type of the value in all circumstances (`SQLITE_TEXT` in this case). However, `sqlite3_column_decltype` will only tell you the declared type of the column, and therefore: (1) is generally meaningless, because SQLite doesn't enforce it at all; and (2) gives back nothing when the column in question was not directly selected from a table column.

> If you want to change default behavior, please add struct field in SQLiteDriver to make opt-in.

Again, I consider the present behavior to be a bug. I don't know why someone would run a query like the above and expect it to come out as a `[]byte`. The [Go 1 compatibility guidelines](https://golang.org/doc/go1compat#expectations) do allow for breaking changes for the sake of fixing bugs. As the existing behavior is surprising and unexpected, forcing people to opt into the correct behavior seems counterproductive. If anything, we should force clients to opt into the legacy behavior if it's really necessary.

mattn:
I don't say I will not change this. Just I didn't understand why current implementation did not work until now. Now, I had  understood why didn't. So I have to add implementation for conversion for sqlite3_column_type except `SQLITE_TEXT` ?

mattn:
addressed in https://github.com/mattn/go-sqlite3/pull/692 with test.

mattn:
Closed by #692

PR #692, opened 2019-02-10
Fix 688

mattn:


rittneje:
@mattn In your test case, can you do `t.Fatalf("non-column string must return string, but got %T", x)` instead? That way if it does fail, we'll know what it got.

PR #693, opened 2019-02-10
Unify travis runs

egonelbre:
This should cut a significant amount of time off the final travis run. See https://travis-ci.com/egonelbre/go-sqlite3/builds/100368565 for proof of concept.

Running time 31.5min
Total time 1h8min

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ before_install:
   - go get golang.org/x/tools/cmd/cover
 
 script:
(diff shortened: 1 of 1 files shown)


rittneje:
@egonelbre Looks like you forgot about the `go build` line in the script, as it still references the `GOTAGS` environment variable. (Also, while you are changing it, you should be able to remove the `GOOS=$(go env GOOS) GOARCH=$(go env GOARCH)` stuff since that's what Go will do by default.)

egonelbre:
@rittneje ah, right... fixed

mattn:
Thank you

PR #734, opened 2019-08-19
darwin/libsqlite3: Also use the homebrew include path

gwd:
When building on darwin with the `libsqlite3` tag, go-sqlite3 adds the
homebrew library path.  It does not, however, add the homebrew include
path, which means that the MacOS sqlite3 header is used instead.  On
my system, this results in build errors that look like this:

./sqlite3_load_extension.go:25:8: could not determine kind of name for C.sqlite3_enable_load_extension
./sqlite3_load_extension.go:33:8: could not determine kind of name for C.sqlite3_load_extension

Add the homebrew include path as well, so that he header matches the
libraries we're using.

Signed-off-by: George Dunlap <george.dunlap@citrix.com>


mattn:
Thank you

PR #744, opened 2019-09-06
Fix context cancellation racy handling

azavorotnii:
[why]
Context cancellation goroutine is not in sync with Next() method lifetime.
It leads to sql.ErrNoRows instead of context.Canceled often (easy to reproduce).
It leads to interruption of next query executed on same connection (harder to reproduce).

[how]
Do query in goroutine, wait when interruption done.

[testing]
Add unit test that reproduces error cases.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -328,7 +328,7 @@ type SQLiteRows struct {
 	decltype []string
 	cls      bool
 	closed   bool
-	done     chan struct{}
+	ctx      context.Context // no better alternative to pass context into Next() method
 }
 
 type functionInfo struct {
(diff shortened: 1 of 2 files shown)


azavorotnii:
will need to check this:
```
=== RUN   TestQueryRowContextCancel
--- FAIL: TestQueryRowContextCancel (4.86s)
	sqlite3_go18_test.go:164: sql: Rows are closed 1
```
wondering if that something different in 1.9 and 1.10 that I didn't assume.

mattn:
Sorry about delay. I'll look into it in later.

mattn:
Thank you

PR #747, opened 2019-09-23
Fix Open() journal mode regression

azavorotnii:
[why]
    see https://github.com/mattn/go-sqlite3/issues/607
    
    SQLite default journal mode is DELETE, but forcing it on open causes "database is locked"
    if other connection exists with WAL mode, for example.
    
    [how]
    Don't set DELETE mode if not set in DSN explicitly.
    
    [testing]
    Run tests in my project where WAL mode is used.


azavorotnii:
@mattn, @GJRTimmer "Coverage decreased (-0.01%) to 50.369%" is not true :)

mattn:
LGTM. Thank you.

PR #754, opened 2019-10-31
Improve readability of Backup()

codesoap:
When reading about the function on godoc.org, I was confused about which connection was the destination and which the source and had to look at the code to find out.


mattn:
Thank you

Issue #764, opened 2019-12-13
Error "cannot start a transaction within a transaction" on cached connection


azavorotnii:
Assuming we run this piece of code:
```
			ctx, cancel := context.WithCancel(ctx)
			tx, err := db.BeginTx(ctx, nil)
```
it is possible to get at the same time context.Cancelled error and started transaction. This will happen only if we cancel context very fast.

This happens because context cancellation is inherently racy as performed not in same go-routine where statement is executed.

rittneje:
@azavorotnii If I understand your bug report correctly, this is not a bug with `BeginTx` specifically, but rather some faulty logic in `(*SQLiteStmt).exec`.
https://github.com/mattn/go-sqlite3/blob/590d44c02bca83987d23f6eab75e6d0ddf95f644/sqlite3.go#L1893-L1905

The existing code erroneously concludes that if we called `sqlite3_interrupt`, then the pending operation will definitely fail so it doesn't bother looking at the result. This is flawed (1) because of the inherent raciness, and (2) because the SQLite docs themselves say that `sqlite3_interrupt` will not necessarily interrupt a pending operation. "If an SQL operation is very nearly finished at the time when sqlite3_interrupt() is called, then it might not have an opportunity to be interrupted and might continue to completion." https://www.sqlite.org/c3ref/interrupt.html

I believe the correct fix would be to not just discard what gets pulled from `resultCh` and instead only return `ctx.Err()` if we receive `SQLITE_INTERRUPT` from the actual operation.

I have also sent a question to the SQLite mailing list about the behavior when `BEGIN` gets interrupted to see if we need to explicitly roll back in that case or not.

azavorotnii:
> I believe the correct fix would be to not just discard what gets pulled from resultCh and instead only return ctx.Err() if we receive SQLITE_INTERRUPT from the actual operation.

@rittneje that was my initial thought, but for some reason instead of SQLITE_INTERRUPT I got SQLITE_DONE. That could be race in my test setup though. Let me check again.

azavorotnii:
@rittneje done as you proposed. I did mistake in my initial test, now it should be stable.

mattn:
Fixed by PR #765 (commit 862b959): exec now checks isInterruptErr before returning ctx.Err(), so a completed-but-cancelled BEGIN no longer leaves a dangling transaction. Closing as resolved.

PR #765, opened 2019-12-13
Fix "cannot start a transaction within a transaction" issue (#764)

azavorotnii:
[why]
If db.BeginTx(ctx, nil) context is cancelled too fast, "BEGIN" statement can be
completed inside DB, but we still try to cancel it with sqlite3_interrupt.
In such case we get context.Cancelled or context.DeadlineExceeded from exec(),
but operation really completed. Connection returned into pool, and returns "cannot
start a transaction within a transaction" error for next db.BeginTx() call.

[how]
If we get context cancelled on "BEGIN" statement, call "ROLLBACK" to clean-up
connection state. Don't return cancellation error from exec() if operation completed
without sqlite3_interrupt.

[testing]
Added unit-test which reproduces issue.

### DIFF
--- a/sqlite3_go18_test.go
+++ b/sqlite3_go18_test.go
@@ -136,6 +136,44 @@ func TestShortTimeout(t *testing.T) {
 	}
 }
 
+func TestExecContextCancel(t *testing.T) {
+	srcTempFilename := TempFilename(t)
+	defer os.Remove(srcTempFilename)
+
+	db, err := sql.Open("sqlite3", srcTempFilename)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer db.Close()
+
+	ts := time.Now()
+	initDatabase(t, db, 1000)
+	spent := time.Since(ts)
+	if spent < 100*time.Millisecond {
+		t.Skip("test will be too racy, as ExecContext below will be too fast.")
+	}
+
+	// expected to be extremely slow query
+	q := `
+INSERT INTO test_table (key1, key_id, key2, key3, key4, key5, key6, data)
+SELECT t1.key1 || t2.key1, t1.key_id || t2.key_id, t1.key2 || t2.key2, t1.key3 || t2.key3, t1.key4 || t2.key4, t1.key5 || t2.key5, t1.key6 || t2.key6, t1.data || t2.data
+FROM test_table t1 LEFT OUTER JOIN test_table t2`
+	// expect query above take ~ same time as setup above
+	ctx, cancel := context.WithTimeout(context.Background(), spent/2)
+	defer cancel()
+	ts = time.Now()
+	r, err := db.ExecContext(ctx, q)
+	// racy check
+	if r != nil {
+		n, err := r.RowsAffected()
+		t.Log(n, err, time.Since(ts))
+	}
+	if err != context.DeadlineExceeded {
+		t.Fatal(err, ctx.Err())
+	}
+}
+
 func TestQueryRowContextCancel(t *testing.T) {
 	srcTempFilename := TempFilename(t)
 	defer os.Remove(srcTempFilename)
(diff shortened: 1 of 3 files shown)


mattn:
Thanks your contribution.

@rittneje Could you please give me comment about this?

azavorotnii:
Doesn't look related to changes in PR at all:
```
The command "$HOME/gopath/bin/goveralls -repotoken 3qJVUE0iQwqnCbmNcDsjYu1nh4J4KIFXx" exited with 1.
```

azavorotnii:
@mattn @rittneje ping here.

sqlite3_go18_test.go
```diff
@@ -136,6 +136,89 @@ func TestShortTimeout(t *testing.T) {
...
+	wg := sync.WaitGroup{}
+	// create several go-routines to expose racy issue
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			for i := 0; i < 1000; i++ {
+				ctx, cancel := context.WithCancel(context.Background())
+				go cancel() // make it cancel concurrently with exec("BEGIN");
```
mattn: I wonder this canceliing always works as you expected. Sleep is needed?

rittneje:
Your implementation looks good. Unfortunately, I cannot think of a great way of dealing with the test deterministically.

sqlite3_go18_test.go
```diff
@@ -136,6 +136,89 @@ func TestShortTimeout(t *testing.T) {
...
+
+	db.SetMaxOpenConns(10)
+	db.SetMaxIdleConns(5)
+
+	defer db.Close()
+	initDatabase(t, db, 100)
+
+	wg := sync.WaitGroup{}
+	// create several go-routines to expose racy issue
+	for i := 0; i < 10; i++ {
```
rittneje: What is the need for this outer for loop? Why is the inner one insufficient?

sqlite3_go18_test.go
rittneje: Unfortunately, the code being tested here is inherently non-deterministic. @azavorotnii Did this test as written fail consistently prior to your changes?

sqlite3_go18_test.go
azavorotnii: inner loop is inside go-routine

sqlite3_go18_test.go
azavorotnii: yes, it failed every time I tested it without fix in code, but I agree that "passed" status doesn't prove issue absence.

sqlite3_go18_test.go
rittneje: Hmm, the only thing I can think of is that the extra goroutines cause the scheduler to behave slightly differently. For example, if it causes the `cancel` goroutine to consistently complete before the `execSync` goroutine, that would explain the discrepancy. (The old code would have consistently hit the "no need to interrupt" case.)

That does give me an idea for something more deterministic here. If we were to call `BeginTx` directly on `*SQliteConn` (rather than going through the database/sql package) with an already canceled context, that would have almost always caused the failure in the old code, but now will always work as expected.

azavorotnii: I did as you suggested and it really much more deterministic (still not 100% as we want to interrupt statement with sqlite3_interrupt.

From tests I see that Conn.Raw() method was introduced only in go 1.13. Will move that test separately.

azavorotnii:
@mattn , @rittneje ping here again.

mattn:
@rittneje any thought?

rittneje:
Sorry for the delay. Looks good to me.

azavorotnii:
@mattn can it be merged now?

djoyner:
I stumbled on this same issue and created a small reproduce (https://gist.github.com/djoyner/0133a23000d3ebd3b5421f975c4c2fbb). FWIW, this PR works.

Kegsay:
Also encountering this problem, it looks like this PR has passed code review so please can it be merged so we can benefit from it downstream without having to fork?

mattn:
will look this in later.

mattn:
Thank you.

PR #772, opened 2020-01-07
added missing mode functionality

AshfordN:
the mode option in the dsn did not work as expected, particularly when trying to open a database in read-only mode

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1000,6 +1000,7 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 	var pkey string
 
 	// Options
+	mode := C.int(C.SQLITE_OPEN_READWRITE | C.SQLITE_OPEN_CREATE)
 	var loc *time.Location
 	authCreate := false
 	authUser := ""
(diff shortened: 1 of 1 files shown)


rittneje:
@AshfordN Do you have an example of it not working correctly? The mode parameter is handled by SQLite itself and is basically applied on top of `SQLITE_OPEN_READWRITE`/`SQLITE_OPEN_READONLY`/`SQLITE_OPEN_CREATE`. Therefore, this change is not needed.

AshfordN:
Well the database isn't actually read-only when you specify mode=ro

rittneje:
Please provide an example code segment where it does not work as expected.

AshfordN:
The following code successfully inserts and updates a record in the database, using go1.13.1 on linux/amd64:

```
package main

import (
	"database/sql"
	"fmt"

)

func main() {
	//open database
	db, err := sql.Open("sqlite3", "main.db?mode=ro")
	if err != nil {
		panic(err)
	}

	//attempt insert
	if _, err := db.Exec("INSERT INTO test(col1, col2) VALUES(1, 2)"); err != nil {
		fmt.Printf("Insert failed: %w\n", err)
	}

	//attempt update
	if _, err := db.Exec("UPDATE test SET col2 = 4 WHERE col1 = 1"); err != nil {
		fmt.Printf("Update failed: %w\n", err)
	}
}
```

rittneje:
See https://www.sqlite.org/uri.html. It should be `sql.Open("sqlite3", "file:main.db?mode=ro")`.

AshfordN:
Ok, this works, thanks.

PR #781, opened 2020-02-04
feat: add go modules

akkumar:
Add go.mod and go.sum to help support go modules ( and remove +incompatible issue in dependent projects ). 

I had chosen golang 1.10 version far enough in the past to support more golang versions but recent enough to get the bugfixes and the like.

Looking forward to getting this in.


mattn:
Thank you

Issue #782, opened 2020-02-08
Making cgo-based callbacks?


gwd:
I've got a callback function which, in one of my key queries, is called thousands of times.  Initially I did the "easy" thing of making a Golang-based callback, using `conn.RegisterFunc()`.  Profiling revealed that queries were spending the vast majority of their time in the marshalling code.

I recently did some benchmark testing, and found that I can get a 20x speedup in this function by writing it in C, and loading it as an extension.  (Benchmarking fingered the golang-based callback as taking nearly 1600ns, and the C-based callback as taking only 80ns.)

However, actually making a separate dynamically-loadable executable is a bit inconvenient for a Go project; I'd have to switch from the built-in Golang build system to a makefile-based system or something.  That's a pain, particularly when I just need to define 1 C function and call 1 C function.

What I'd really like to be able to do is something like this:
```golang
/*
[Define 'myFunc']
*/
import "C"

/* ... */

func myFuncHook(conn *sqlite3.SQLiteConn) error {
   db := conn.GetContext() // Return underlying SQLiteContext
  ret = C.sqlite3_create_function(db, "myFunc", 3, SQLITE_UTF8, NULL, C.myFunc /* ... */)
  /* ... */
}
```

And then set `myFuncHook` as the `ConnectionHooks`.  I see that the underlying C SQLite connection has its own type, but I don't see how to get the connection from the hook callback.

Is that possible?  And if it's not possible, could it be made possible?

Alternately, would it make sense to add a `RegisterCFunc` method, which would take a C function pointer?

I might be able to send you a PR if I knew what kind of solution you'd prefer.

Thanks in advance!

gwd:
(Alternately, if you think the Golang callback could be sped up significantly, I could post the function and my simple benchmarking code; but my impression was that there was no way to retain the type safety properties without all the overhead of reflection in the unmarshal / marshal paths.)

rittneje:
https://golang.org/cmd/cgo/
> Cgo translates C types into equivalent unexported Go types. Because the translations are unexported, a Go package should not expose C types in its exported API: a C type used in one Go package is different from the same C type used in another.

I don't think we can add something like `GetContext() *C.sqlite3` as you proposed. However, I believe something like this could work:

```go
func (c *SQLiteConn) Raw(f func(unsafe.Pointer) error) error {
    c.mu.Lock()
    defer c.mu.Unlock()
    if err := f(unsafe.Pointer(c.db)); err != nil {
[...]

PR #784, opened 2020-02-08
SQLiteConn: Add 'Raw' method to give access to underlying sqlite3 con…

gwd:
…text structure

This is necessary, for instance, to allow users of the package to make
their own CGo callbacks, and have the underlying sqlite3 library call
those C functions directly.

Note that the intention from the sqlite3 library seems to be that
there should be a measure of binary compatibility between versions:
even extensions compiled against an older different version of sqlite3
should be able to call sqlite3_create_function with no problems. So
this should work even if users of the package have a slightly
different version of the sqlite3.h header.

This fixes https://github.com/mattn/go-sqlite3/issues/782 .

Signed-off-by: George Dunlap <dunlapg@umich.edu>

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -108,5 +108,12 @@ call RegisterFunction from ConnectHook.
 
 See the documentation of RegisterFunc for more details.
 
+Cgo SQLite3 Extensions
+
+Go callbacks are convenient, but the runtime overhead of reflecting
+incoming types can be significant.  For performance-critical
+functions, Cgo functions can also be defined.  See SQLiteConn's Raw
+method for an example.
+
 */
 package sqlite3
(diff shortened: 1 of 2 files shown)


gwd:
@rittneje FYI

gwd:
Gah -- somehow I missed the `c.mu.Lock()` call in @rittneje 's example.  I'll wait for any other comments and then send v2.

rittneje:
@gwd I would suggest doing something like the typecheck for `ErrNo` that I had in my example. That way clients can easily leverage the existing logic for constructing an `Error` object for the error code, extended error code, etc.

zepatrik:
For anyone following this issue: new PR #785

Issue #798, opened 2020-04-02
Error loading CSV extension


demontiejr:
Hi,

I'm trying to use the CSV extension but I'm getting a "not an error" message whenever I try to execute a statement.

I was able to compile the [csv.c](https://www.sqlite.org/src/raw/ext/misc/csv.c?name=531a46cbad789fca0aa9db69a0e6c8ac9e68767d) file by running `gcc -g -fPIC -shared csv.c -o csv.so -lsqlite3 -lpcre` (I'm using MacOS X, so it's actually clang). I could load the extension from the SQLite3 client and create the virtual table. However, when I try to load the extension, I'm getting "not an error" error message.

```go
	sql.Register("sqlite3_with_extensions",
		&sqlite3.SQLiteDriver{
			Extensions: []string{
				"csv",
			},
		})
```

I then cloned the go-sqlite3 repository and tried to execute some of the examples (such as `mod_regex`), but I'm getting the same error. I'm using `CGO_ENABLED=1`, so the problem is not related to it. Does someone know how to get this working?

Thanks,
Demontiê

rittneje:
https://github.com/mattn/go-sqlite3/blob/67986a7832dcbda29de5c1341b3dec6c6d97511a/sqlite3_load_extension.go#L33-L37

I'm guessing that `sqlite3_load_extension` failed, but then because `sqlite3_enable_load_extension` got called, `sqlite3_errmsg` is actually returning the "error message" corresponding to `SQLITE_OK`.

demontiejr:
Hi rittneje,

thanks for the reply. I do believe the call to sqlite3_load_extension failed
but I have no idea why, since I can load the extension from the CLI. Do you
have a guess?

Em qui, 2 de abr de 2020 19:24, rittneje <notifications@github.com>
escreveu:

rittneje:
I'm not very familiar with SQLite extension I'm afraid. One thing to try is to make sure that the `LD_LIBRARY_PATH` environment variable (or `DYLD_LIBRARY_PATH` on Mac) includes the directory in which the shared library is located.

By the way, the command you posted above built sqlite3_mod_regexp.so and not csv.so. Was that a typo?

demontiejr:
I tried setting `DYLD_LIBRARY_PATH` but it didn't help. The weird thing is that the same error is happening when I try to load any extension (i.e., when I build the examples with `make` and try to execute them).

[...]

PR #800, opened 2020-04-04
report actual error message if sqlite3_load_extension fails

rittneje:
Apparently if `sqlite3_load_extension` fails, it will not set the error to be returned by `sqlite3_errmsg`. This was the cause of confusion in #798.


mattn:
Thank you

Issue #384, opened 2017-02-21
Cross compiling fails


SamHH:
Very possibly a mistake on my part, if so please let me know where I went wrong. My first time with Go today!

I have a basic Go file using this library. I am on macOS `10.12.3`, Go `1.8 darwin/amd64`.

Succeeds: `env GOOS=darwin go build -i app.go`

Fails: `env GOOS=linux go build -i app.go`

Specifying `GOARCH=amd64` makes no difference.

Errors:
```
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:18: undefined: SQLiteConn
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:26: undefined: SQLiteConn
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:27: undefined: namedValue
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:29: undefined: namedValue
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:35: undefined: SQLiteConn
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:36: undefined: namedValue
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:44: undefined: SQLiteConn
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:49: undefined: SQLiteConn
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:54: undefined: SQLiteStmt
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:63: undefined: SQLiteStmt
../../../go/src/github.com/mattn/go-sqlite3/sqlite3_go18.go:36: too many errors
```

Said very basic Go file causing this, in-case it helps:

```
package main

import (
	"database/sql"
	"log"
	"os"
	_ "github.com/mattn/go-sqlite3"
)

// Bookmark represents a single row from database
type Bookmark struct {
	Id       int
	Url      string
	Metadata string
	Tags     string
	Desc     string
	Flags    int
}

func main() {
	dbpath := GetDbPath()

	db := InitDB(dbpath)
	defer db.Close()

	bookmarks := GetAllBookmarks(db)
	log.Print(bookmarks)
}

// Write errors to log & quit
func CheckError(err error) {
	if err != nil {
		log.Fatal(err)
	}
}

[...]

PR #804, opened 2020-04-17
 Document requirements for cross compiling from OSX

vishnukvmd:
Inspiration: https://github.com/mattn/go-sqlite3/issues/384#issuecomment-433584967

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,25 +213,14 @@ This library can be cross-compiled.
 In some cases you are required to the `CC` environment variable with the cross compiler.
 
 ## Cross Compiling from MAC OSX
-Install the `musl-cross` and `mingw-w64` formulae.
-```bash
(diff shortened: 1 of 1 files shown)


gertcuykens:
Don't think it's complete? One is for compiling to alpine (musl) other for compiling to windows and non of them compile to Linux (libc debian)

Also I believe the `-ldflags "-linkmode external -extldflags -static"` are important to explain?

vishnukvmd:
@gertcuykens I ran into more issues while cross compiling for ARM, and at this point I'd heavily recommend using [xgo](https://github.com/karalabe/xgo/) which seems to work out of the box for a [wide variety of targets](https://github.com/karalabe/xgo/blob/master/README.md#limit-build-targets).

mattn:
Thank you

PR #805, opened 2020-04-21
Fix for 'regex' early termination of returned rows

ShanePerron:
Once the regex encountered the first instance of a non-match, it would return without processing the rest of the rows in the statement. This change allows it to process the remaining, only setting the sqlite3_result_int to zero then continuing. This worked fine for the example as it only had one item to process.


mattn:
Thank you.

Issue #499, opened 2017-11-29
prepared multiple statement queries


temoto:
Hello. Thanks for your work on go-sqlite3.

I'm having trouble with this prepared multi-statement query.

```go
	if self.pushStmt, err = self.db.PrepareContext(ctx,
		`update queue
			set modified = @modified, scheduled = @scheduled where id = @id;
		insert into queue (id, created, modified, scheduled, data)
			select @id, @created, @modified, @scheduled, @data
			where (select Changes() = 0);
		`); err != nil {
		log.Fatal("bender.Init db prepare push statement error:", err)
	}
// ...
	_, err := self.pushStmt.QueryContext(ctx,
		sql.Named("id", qi.Id),
                //...
)
```

> error: sql: statement expects 3 inputs; got 5

3 inputs is count of placeholders in first statement (update).

Without prepare, using `ExecContext` gives another error:

> column index out of range

Switching to numbered placeholder syntax `?1, ?2...` gives another error:

> not enough args to execute query: want 9 got 5

Switching to position dependent placeholder syntax `?` - works fine. But I don't want to create pain for future readers of that code.

Any help is much appreciated.

vith:
This just got me too. The `database/sql` docs give examples that use `@` as the prefix for named placeholders: https://golang.org/pkg/database/sql/#Named

But `go-sqlite3` uses `:` as the prefix: https://github.com/mattn/go-sqlite3/blob/25ecb14adfc7543176f7d85291ec7dba82c6f7e4/sqlite3_go18_test.go#L36

So arguably there are three problems:
1. Not implementing what the go stdlib sql interface docs describe
2. The difference seems to be undocumented
3. The error is misleading, since the problem has nothing to do with columns?

rittneje:
@vith I have filed #697 to resolve a few issues with how this library handles named parameters. Please note that the format of named parameters is defined by the actual driver (so SQLite itself in this case, not this library).

emersion:
It seems this has been fixed in #811. Maybe we can close this issue?

Issue #697, opened 2019-02-27
named parameters not handled correctly


rittneje:
As per [the documentation](https://www.sqlite.org/c3ref/bind_blob.html), SQLite allows three types of parameter names: `@foo`, `:foo`, and `$foo`. However, this library only considers the `:foo` form, which causes the following perfectly valid query to fail:
```go
db.Exec("INSERT INTO Foo VALUES (@fooid)", sql.Named("fooid", 5))
```

Furthermore, this library does not properly handle the user providing a parameter with an unrecognized name, as can be seen [here](https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L1777). If the parameter name is unrecognized, `sqlite3_bind_parameter_index` returns 0. But this return value is blindly passed to the various `sqlite3_bind_X` routines. These routines are expecting an index of at least 1, so such a call is plainly invalid. Parameters with unrecognized names should be ignored.
```go
db.Exec("INSERT INTO Foo VALUES (:fooid)", sql.Named("fooid", 5). sql.Named("barid", 10))
```
This tries to bind 10 at index 0 (since `barid` isn't a valid parameter). Instead, the `barid` parameter should be ignored.

emersion:
It seems this has been fixed in #811. Maybe we can close this issue?

PR #806, opened 2020-04-22
Enable all prefixes for named parameters and allow for unused named parameters 

gber:
Particularly "@"-based named parameters should be supported since they are mentioned in the Go documentation. This fixes issues #499 and #697.


gber:
Superseded by PR #811

PR #809, opened 2020-05-03
Fix #808: remove goconvey

### LINKED ISSUE #808: smartystreets/goconvey now requiring go 1.13
Pipelines are now failing because we can't `go get github.com/smartystreets/goconvey` any longer on go version < 1.13. Would need to pin versions or use other tricks to make things compatible, however goconvey doesn't seem to be used any longer since #592, so might as well just remove that 'go get' from the CI job.

### LINKED ISSUE #810: Inconsistent test coverage reports
Looks like this is happening: https://github.com/lemurheavy/coveralls-public/issues/1182

Should be a simple matter of synchronising coveralls with the travis job using the parallel flag and notifications. Might as well add the fix to #809.

itizir:
Closes #808.
Closes #810.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ before_install:
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew update
     fi
-  - go get github.com/smartystreets/goconvey
   - go get github.com/mattn/goveralls
   - go get golang.org/x/tools/cmd/cover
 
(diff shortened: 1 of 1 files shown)


itizir:
Hmmm. Not clear to me whether the coveralls change did anything. Did I misunderstand something?

itizir:
Oooh! I get it now: since the latest master commit only ran 4 out of 12 jobs (because of failures due to #808), coveralls computed coverage based on those 4 only, and they happen to have slightly higher coverage... 🤦

mattn:
Thank you

PR #827, opened 2020-07-04
Update "database is locked" documentation

arp242:
This explains a bit more what is going on, removes the recommendation to
set cache to shared (which is considered deprecated), and mentions using
the WAL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,17 +526,23 @@ For an example see [dinedal/go-sqlite3-extension-functions](https://github.com/d
 
 - Error: `database is locked`
 
-    When you get a database is locked. Please use the following options.
+    This means two different connections are trying to write to the database at
+    the same time.
 
-    Add to DSN: `cache=shared`
+    One way to fix this is to ensure your program limits the number of writers
+    to 1, for example by using `sync.Mutex` (it's fine to be have multiple
+    reads, even when writing).
 
-    Example:
+    Adding `_journal_mode=wal` improves the general concurrency in most cases,
+    although it doesn't 100% eliminate locking issues:
+
     ```go
-    db, err := sql.Open("sqlite3", "file:locked.sqlite?cache=shared")
+    db, err := sql.Open("sqlite3", "file:locked.sqlite?_journal_mode=wal")
     ```
 
-    Second please set the database connections of the SQL package to 1.
-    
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -526,17 +526,23 @@ For an example see [dinedal/go-sqlite3-extension-functions](https://github.com/d
 
 - Error: `database is locked`
 
-    When you get a database is locked. Please use the following options.
+    This means two different connections are trying to write to the database at
+    the same time.
 
-    Add to DSN: `cache=shared`
+    One way to fix this is to ensure your program limits the number of writers
+    to 1, for example by using `sync.Mutex` (it's fine to be have multiple
```
rittneje, line 533: Having multiple writers is actually not itself a problem. However, what can cause an issue even with write-ahead logging is if a connection tries to write after reading within a transaction.

```go
tx := database.Begin()
...
tx.QueryRow(`SELECT ID FROM Foo WHERE ...`)
...
tx.Exec(`UPDATE Foo ...`)
...
```

The problem here is that if the database gets modified between the `SELECT` and the `UPDATE` by another connection, there is no way for SQLite to properly retain the "isolation" property of ACID. This is because the interim modification may have changed the very data you read out. This issue can be avoided by always using immediate locking behavior (i.e., `BEGIN IMMEDIATE`) for any transactions that may ultimately write to the database. (It of course is also avoided by throttling the connection pool to one, provided that no other applications will also be writing to the same database.)
```

File: README.md [hunk 2]
```diff
@@ -526,17 +526,23 @@ For an example see [dinedal/go-sqlite3-extension-functions](https://github.com/d
...
 
-    Example:
     ```go
-    db, err := sql.Open("sqlite3", "file:locked.sqlite?cache=shared")
+    db, err := sql.Open("sqlite3", "file:locked.sqlite?_journal_mode=wal")
     ```
 
-    Second please set the database connections of the SQL package to 1.
-    
+    To guarantee only one connection is being run at the same time yoi can set
```
tshepang, line 543:
```suggestion
    To guarantee only one connection is being run at the same time, you can set
```

tshepang:
As for me, problem was fixed by closing Rows, as shown on https://stackoverflow.com/a/32483180/321731.

DmitriyMV:
Kindly ping @arp242

PR #828, opened 2020-07-07
Remove dependency on goquery and clear go.mod requires

dolmen:
This is a rewrite of the upgrade/upgrade.go script (which is used only by maintainers) to use only packages from stdlib, and in particular remove the use of the external dependency on `github.com/PuerkitoBio/goquery`.

As a result, `go.mod` is now cleaned from any external dependency and all Go modules that depend on go-sqlite3 will also have the goquery dependency cleaned from their dependency tree (`go.sum`).


dolmen:
@mattn Could you review this? It would seriously help downstream users to manage their dependencies...

dolmen:
Rebased.

mattn:
Sorry my delay and thank your contribution. But I hope to use github.com/PuerkitoBio/goquery to make this easily. So I'll merge my PR just adding ignore build tag.

mattn:
Thanks.

dolmen:
So you prefer to use an external package and more lines of code that just a regexp? 
😞

mattn:
Now I exported upgrade package to https://github.com/mattn/go-sqlite3-upgrade

dolmen:
You could also have just created a separate Go module in the `upgrade` directory. It make sense to keep the upgrade script in the same repository.

PR #829, opened 2020-07-07
Travis-CI: set go_import_path

dolmen:
Set `go_import_path` to tell Travis-CI where to checkout the code, so the Travis build can also work on forks. This is important when building without Go modules support.


mattn:
Thank you

PR #830, opened 2020-07-07
Travis-CI: add Go 1.14.x, change order of Go versions

dolmen:
Changes in Go versions for the Travis-CI builds:
- reverse order of Go versions so the latest ones are checked first (we get the most interesting results earlier)
- add Go 1.14.x (latest stable) at the first position
- use 'tip' instead of 'master' (like https://tip.golang.org) and move it just after 1.14.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,13 @@ addons:
     update: true
 
 go:
-  - 1.9.x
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

PR #848, opened 2020-09-01
RowsColumnTypeNullable not implemented

jinzhu:
Hello @mattn 

When I am trying the `columnType`'s `Nullable` method, seems it always returns `true` although it is not implemented yet, maybe we should comment out this method right now or always return `false`, `false`?

Thank you.


jinzhu:
Seems the CI error not related to the PR change

mattn:
Thank you

PR #861, opened 2020-10-02
Fix #860 extenstion entry bug

### LINKED ISSUE #860: Entrypoint ignored when loading an extension
When loading an extension with a custom entrypoint the default entrypoint is called.

```go
sql.Register("sqlite3-crypt", &sqlite3.SQLiteDriver{
		// always calls the default entrypoint and not init_db
		return conn.LoadExtension("/path/to/ysqlite3_crypt_vfs.ext", "init_db")
	},
})
```

The extension:
```cpp
#ifdef _WIN32
__declspec(dllexport)
#endif
    extern "C" int init_db(sqlite3* db, char** error_message, const sqlite3_api_routines* api) noexcept
{
	/* I want to call this entrypoint */
}

#ifdef _WIN32
__declspec(dllexport)
#endif
    extern "C" int sqlite3_ysqlitecryptvfs_init(sqlite3* db, char** error_message,
                                                const sqlite3_api_routines* api) noexcept
{
	/* But this entrypoint is always called */
}
```

terrakuh:
Fix overshadowing of entrypoint variable.

### DIFF
--- a/sqlite3_load_extension.go
+++ b/sqlite3_load_extension.go
@@ -68,7 +68,7 @@ func (c *SQLiteConn) loadExtension(lib string, entry *string) error {
 
 	var centry *C.char
 	if entry != nil {
-		centry := C.CString(*entry)
+		centry = C.CString(*entry)
 		defer C.free(unsafe.Pointer(centry))
 	}
 


mattn:
Right. Thank you.

PR #865, opened 2020-10-13
Fix racy test

johejo:


evanj:
I haven't tested if this PR fixes the problem, but I can confirm that `TestExecContextCancel` is "flaky": it frequently fails where a retry will fix the problem. It seems to me like the one line change to the `_test.go` file might be the fix that should be merged, but I haven't looked at this in detail.

johejo:
That point is correct.  This PR is not suitable.

PR #872, opened 2020-11-10
sqlite3.go: Remove -DSQLITE_ENABLE_FTS4_UNICODE61: not supported

evanj:
This option was enabled by default in sqlite3 on 2014-07-03.
This setting does nothing. It can now be disabled with
SQLITE_DISABLE_FTS3_UNICODE. See the upstream commit:
https://sqlite.org/src/info/0cc0230ae9cfc976

I think this change was imported into this project with commit
ee9da4840dd680d1fe2927692ea4ab5f8cd7c91c on 2015-06-12.


mattn:
Thanks!

PR #877, opened 2020-11-13
Support vfs for Open

### LINKED ISSUE #876: Support VFS when opening a db
Hi,

is there a reason why it is currently not supported to define a VFS via the connection string when opening a DB?

I guess everything is prepared for it but the `zVfs` parameter is just not use and always set to `nil`
https://github.com/mattn/go-sqlite3/blob/92f580b350d88e068bfbe223ea51f64044c0f280/sqlite3.go#L1377

Would you mind to add the vfs to the connection uri parameters, so that e.g. this can be done:

```
db, err := sql.Open("sqlite3", "test.db?cache=shared&mode=rwc&vfs=unix-dotfile")
```

(I guess as a start, it is good enough only supporting the build-in-VFS'; registering new VFS' can be added if needed later)

mattn:
Closes #876

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1041,6 +1041,7 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 	secureDelete := "DEFAULT"
 	synchronousMode := "NORMAL"
 	writableSchema := -1
+	vfsName := ""
 
 	pos := strings.IndexRune(dsn, '?')
 	if pos >= 1 {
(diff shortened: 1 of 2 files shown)


guenhter:
Thx. Tested it. Works like a charm

jtarchie:
@guenhter, are you able to clarify how you can load your VFS? I'm unable to use this load extension before `sqlite_open`.

guenhter:
@jtarchie sorry, I'm not working for for that company any more and my brain seems not to be able to remember the stuff :)

Issue #386, opened 2017-03-01
Compiling on Darwin no longer works


otoolep:
Something has changed here, I can no longer compile [rqlite](https://github.com/rqlite/rqlite) on Darwin (MacOSX):
```
# github.com/mattn/go-sqlite3
../../mattn/go-sqlite3/sqlite3-binding.c:28837:10: error: use of unknown builtin '__builtin_add_overflow' [-Wimplicit-function-declaration]
../../mattn/go-sqlite3/sqlite3-binding.c:28857:10: error: use of unknown builtin '__builtin_sub_overflow' [-Wimplicit-function-declaration]
../../mattn/go-sqlite3/sqlite3-binding.c:28857:10: note: did you mean '__builtin_add_overflow'?
../../mattn/go-sqlite3/sqlite3-binding.c:28837:10: note: '__builtin_add_overflow' declared here
../../mattn/go-sqlite3/sqlite3-binding.c:28872:10: error: use of unknown builtin '__builtin_mul_overflow' [-Wimplicit-function-declaration]
../../mattn/go-sqlite3/sqlite3-binding.c:28872:10: note: did you mean '__builtin_sub_overflow'?
../../mattn/go-sqlite3/sqlite3-binding.c:28857:10: note: '__builtin_sub_overflow' declared here
```

This was fine a couple of weeks back.

otoolep:
I'm running OSX El Capitain, version 10.11.14.
```
$ gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 7.0.2 (clang-700.1.81)
Target: x86_64-apple-darwin15.4.0
Thread model: posix
```

mattn:
```c
#include <stdio.h>
# define GCC_VERSION (__GNUC__*1000000+__GNUC_MINOR__*1000+__GNUC_PATCHLEVEL__)

int
main(int argc, char* argv[]) {
  printf("%d\n", GCC_VERSION);
  return 0;
}
```
What value do you get?

otoolep:
```
~/i $ cat main.c 
#include <stdio.h>
# define GCC_VERSION (__GNUC__*1000000+__GNUC_MINOR__*1000+__GNUC_PATCHLEVEL__)

int
main(int argc, char* argv[]) {
  printf("%d\n", GCC_VERSION);
  return 0;
}
~/i $ gcc main.c 
[...]

PR #878, opened 2020-11-13
sqlite3.go: remove -DSQLITE_DISABLE_INTRINSIC: better builds

evanj:
This "disables the use of compiler-specific built-in functions such
as __builtin_bswap32()" (from the SQLite docs) so this change might
produce slightly better code. My primary motivation, however, is that
the "default" configuration for SQLite, which is widely tested, does
not set this preprocessor macro.

From looking at Github issues, it appears this was added to avoid a
build error on Mac OS X 10.11, in 2017:
https://github.com/mattn/go-sqlite3/issues/386

There have been a number of changes to sqlite3 since then.
I found a machine running Mac OS X 10.11.6. It was able to build and
run the tests in this package with this change.

Mac OS X 10.11 is has not been supported by Apple since 2018
(currently Apple is releasing updates for Mac OS 10.13 and newer; 11
is the current release). However, Go 1.14 is supported, and it
requires Mac OS X 10.11 or newer: https://golang.org/doc/go1.14
Go 1.15 only supports Mac OS 10.12 and newer:
https://golang.org/doc/go1.15


mattn:
Good to hear.

> I found a machine running Mac OS X 10.11.6. It was able to build and
run the tests in this package with this change.

Thank you.

PR #879, opened 2020-11-15
TestExecContextCancel: Reduce timeout to make less flaky

evanj:
This test fails fairly often. On my system, I can trigger it with:

    go test . -run=TestExecContextCancel -count=10 -failfast -v

This makes the test less flaky by timing out the context after a
consistent 50 millisecond delay. This was enough time for the query
to start, then get cancelled with sqlite3_interrupt() in my tests.
This now passes the above check.

This is a modified version of the change suggested in:
https://github.com/mattn/go-sqlite3/pull/865


mattn:
Thank you this too!

Issue #870, opened 2020-10-30
Creating User-Defined SQLITE3 Golang Function not Working


PoseidonCoder:
```go
//omitting meta imports and package name for brevity

func regex(re, s string) (bool, error) {
	return regexp.MatchString(re, s)
}

func check(err error) {
	if err != nil {
		panic(err)
	}
}

func main {
    sql.Register("sqlite3_with_go_func",
		&sqlite3.SQLiteDriver{
			ConnectHook: func(conn *sqlite3.SQLiteConn) error {
				return conn.RegisterFunc("regexp", regex, true)
			},
		})

   db, err := sql.Open("sqlite3", "./Dictionary.sqlite")
   check(err)

    test, err := db.Query("SELECT * FROM entries WHERE word REGEXP 'a' LIMIT 10")
    check(err)
}
```

But when I run it, I get an error saying:
```
panic: no such function: REGEXP
``` 

I use a Windows 10 computer and when I try to define a user-defined function in SQLITE3, I get an error saying that that user-defined function does not exist.

rittneje:
You registered your augmented driver as `"sqlite3_with_go_func"`, so you need to provide that same name to `sql.Open` instead of `"sqlite3"`.

PoseidonCoder:
Thanks! That worked!

PR #880, opened 2020-11-16
Expand documentation for extension functions

macaullyjames:
This relates to #870; it's not immediately clear that you need to pass a
different driver name to sql.Open from the documentation.


mattn:
LGTM. I'll merge after CI passed.

mattn:
Thank you

PR #881, opened 2020-11-16
sqlite3_test.go: Fix go test -run=...: Use standard sub-tests

evanj:
Selecting only some tests with go test -run=... does not work, because
some of the tests are executed using testing.RunTests(). That function
is documented as "an internal function". This changes TestSuite to use
the testing subtests feature instead.

This has a behaviour change: the benchmarks now need to be
selected at the command line with the standard go test -bench=.
flag. This will also set up the test database twice when running
benchmarks, rather than once.

### DIFF
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -26,7 +26,7 @@ import (
 	"time"
 )
 
-func TempFilename(t *testing.T) string {
+func TempFilename(t testing.TB) string {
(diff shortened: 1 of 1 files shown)


mattn:
Thank you. But tests seems failing.

mattn:
Could you please rebase from master ?

mattn:
Updated patch to reduce redrawing of popupmenu

PR #882, opened 2020-11-16
doc.go: you can use Conn.Raw to get *SQLiteConn

evanj:
This can be easier that registering a new driver, in some cases.
Add a test to verify that this works.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -79,9 +79,8 @@ Then, you can use this extension.
 
 Connection Hook
 
-You can hook and inject your code when the connection is established. database/sql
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

mattn:
some tests are failing...

evanj:
I failed to read the documentation carefully enough, sorry. Conn.Raw was added in 1.13. The Go project supports 1.15 and 1.14 right now, so we either need to revert, or decide to only support those versions. I'll prepare the revert PR now.

PR #883, opened 2020-11-16
sqlite3_test.go: Move Go 1.13 test to sqlite3_go113_test.go

evanj:
Commit 4f7abea96e added a test that uses Conn.Raw, which was added in
Go >= 1.13. The go-sqlite3 project runs tests with Go >= 1.11. Remove
the test from sqlite3_test.go, so it only runs with the correct
versions of Go.

Instead of adding a new test, modify the existing test that already
uses Conn.Raw() to check the type of driverConn.

Fixes issue from #882

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -90,7 +90,7 @@ ConnectHook to get the SQLiteConn.
 					},
 			})
 
-You can also use database/sql.Conn.Raw:
+You can also use database/sql.Conn.Raw (Go >= 1.13):
 
 	conn, err := db.Conn(context.Background())
 	// if err != nil { ... }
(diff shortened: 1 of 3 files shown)


mattn:
Thank you! I'll merge after CI pass.

evanj:
The Windows failure is: ```--- FAIL: TestExecContextCancel (57.89s)
    sqlite3_go18_test.go:170: 1000000 <nil> 26.4808527s
    sqlite3_go18_test.go:173: <nil> context deadline exceeded``` which might be helped by #879 . The Mac failure is the one that is always happening on Mac OS X with 1.11 due to some dependency issue that I haven't looked at.

I hope I did this correctly this time! 😄  thank you!

mattn:
Thanks!

PR #885, opened 2020-12-02
add support for defining an "eponymous only" virtual table

### LINKED ISSUE #846: Eponymous-only virtual tables
Does this package support [eponymous-only virtual tables](https://sqlite.org/vtab.html#eponymous_only_virtual_tables)? I see some vtable related files that have the sqlite_vtable and vtable build tags, but I can't tell if they would allow a NULL xCreate.

patrickdevivo:
As suggested here: https://github.com/mattn/go-sqlite3/issues/846#issuecomment-736206222

Ultimately, should close #846 

- [x] Add some test cases (at least one)
- [x] Add an example demonstrating usage

### DIFF
--- a/_example/vtable_eponymous_only/main.go
+++ b/_example/vtable_eponymous_only/main.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	"log"
+
+	"github.com/mattn/go-sqlite3"
+)
+
+func main() {
+	sql.Register("sqlite3_with_extensions", &sqlite3.SQLiteDriver{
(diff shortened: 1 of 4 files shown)


patrickdevivo:
Just wanted to notify @rittneje @mattn, this is a follow-up to #846 that adds support for eponymous only virtual table modules (thank you @rittneje for the instructions/guidance!). Hopefully this is a reasonable addition/change, please let me know - happy to make any amendments!

anacrolix:
Looks great!

mattn:
Looks great. Thanks.

PR #893, opened 2020-12-16
clarify usleep license

bradrydzewski:
This pull request helps clarify the license of the `usleep` function.  This `usleep` function originates from the PHP codebase and is therefore published under an OSI-approved license. However, linking to stackoverflow without linking back to the original source caused some problems during a recent license audit:

> Anything from stackoverflow is under the Creative Commons Attribution ShareAlike 4.0 license which requires that any code that remixes, transforms or builds upon the material on stackoverflow be distributed under the same license. So Go-sqllite3 could potentially be violating stackoverflow TOS.

This change should clarify that this file is open source and does not violate the stackoverflow TOS, and should help avoid any license compliance challenges in the future 🙂

### DIFF
--- a/sqlite3_usleep_windows.go
+++ b/sqlite3_usleep_windows.go
@@ -15,7 +15,9 @@ package sqlite3
 // This code should improve performance on windows because
 // without the presence of usleep SQLite waits 1 second.
 //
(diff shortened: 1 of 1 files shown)


mattn:
Thank you.

PR #894, opened 2020-12-26
Add ?_cache_size=[..] to connection parameters

arp242:
Add a shortcut for PRAGMA cache_size; this is a pretty useful setting:
the default of -2000 (2M) is not especially high, and a lot of people
will probably want to increase this.

For example, while running a bunch of fairy expensive queries in
parallel:

	With SetMaxOpenConns(1):
	 -2000:  5762ms
	-20000:  4714ms

	With SetMaxOpenConns(20):
	 -2000:  3067ms
	-20000:  2532ms

Which isn't a bad performance boost for changing a single number.


mattn:
Thank you

PR #895, opened 2020-12-26
Export sqlite3_stmt_readonly() via SQLiteStmt.Readonly()

arp242:
This can be used like in the test; I wrote a little wrapper around
sql.DB which uses this, and allows concurrent reads but just one single
write. This is perhaps a better generic "table locked"-solution than
setting the connections to 1 and/or cache=shared (although even better
would be to design your app in such a way that this doesn't happpen in
the first place, but even then a little seat belt isn't a bad thing).

The parsing adds about 0.1ms to 0.2ms of overhead in the wrapper, which
isn't too bad (and it caches the results, so only needs to do this
once).

At any rate, I can't really access functions from sqlite3-binding.c from
my application, so expose it via SQLiteStmt.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -2007,6 +2007,13 @@ func (s *SQLiteStmt) execSync(args []namedValue) (driver.Result, error) {
 	return &SQLiteResult{id: int64(rowid), changes: int64(changes)}, nil
 }
 
+// Readonly reports if this statement is considered readonly by SQLite.
+//
+// See: https://sqlite.org/c3ref/stmt_readonly.html
+func (s *SQLiteStmt) Readonly() bool {
+	return C.sqlite3_stmt_readonly(s.s) == 1
+}
+
 // Close the rows.
 func (rc *SQLiteRows) Close() error {
 	rc.s.mu.Lock()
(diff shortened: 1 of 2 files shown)


mattn:
If you want to check the database is readonly:

```go
package main

import (
	"database/sql"
	"fmt"
	"log"

)

func main() {
	db, err := sql.Open("sqlite3", "foo.db?_query_only=true")
	if err != nil {
		log.Fatal(err)
	}
	var readonly bool
	err = db.QueryRow("pragma query_only").Scan(&readonly)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(readonly)
}
```

Do you need to refer to the native interface?

arp242:
That's not quite the same @mattn; what I want to check if a query will write the database (i.e. `insert`), or is just a read operation (i.e. `select`).

Then, if you spawn 100 `select * from ...` goroutines and 100 `insert into ...` goroutines, the reads will be processed in parallel (depending on `SetMaxOpenConns()`), but the inserts will be processed in serial if you use a wrapper which calls this and stops execution with a mutex.

mattn:
Ah, sorry. I had not make sure your description well. I'll merge if tests will be passed.

arp242:
No worries, could have explained it better maybe 😅

It looks like `Conn.Raw()` didn't get added until Go 1.13. This is just just in the test, so I moved the test to `sqlite3_go113_test.go` and now the Go 1.12 test runners pass.

mattn:
Thank you

PR #899, opened 2020-12-30
Make Error work work with and without cgo enabled

arp242:
Let's say you have the following code:

	// ErrUnique reports if this error reports a UNIQUE constraint violation.V
	func ErrUnique(err error) bool {
		var sqlErr *sqlite3.Error
		if errors.As(err, &sqlErr) && sqlErr.ExtendedCode == sqlite3.ErrConstraintUnique {
			return true
		}
		var pqErr *pq.Error
		if errors.As(err, &pqErr) && pqErr.Code == "23505" {
			return true
		}
		return false
	}

This doesn't work with CGO_ENABLED=0 since:

	$ CGO_ENABLED=0 go test .
	# zgo.at/zdb [zgo.at/zdb.test]
	./zdb.go:16:14: undefined: sqlite3.Error
	./zdb.go:17:55: undefined: sqlite3.ErrConstraintUnique

Now, it's not too hard to create two implementations of this: a cgo version
which includes the SQLite and PostgreSQL check, and a non-cgo version which
includes just the PostgreSQL check, but it's all a bit cumbersome and requires
duplicate code or wrapper functions and more indirection than really needed.

Since there are just two methods that actually use the C "package", I think it
makes sense to put those in a error_cgo.go file, and put some basic shims in the
static_mock.go file – this way the code always compiles.

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,25 +1,16 @@
-// Copyright (C) 2019 Yasuhiro Matsumoto <mattn.jp@gmail.com>.
-//
-// Use of this source code is governed by an MIT-style
-// license that can be found in the LICENSE file.
-
 package sqlite3
 
-/*
-#ifndef USE_LIBSQLITE3
-#include <sqlite3-binding.h>
-#else
-#include <sqlite3.h>
-#endif
-*/
-import "C"
 import "syscall"
+
+// Make sure the error interface is implemented.
+var (
+	_ error = Error{}
+	_ error = ErrNo(0)
+	_ error = ErrNoExtended(0)
+)
 
 // ErrNo inherit errno.
 type ErrNo int
-
-// ErrNoMask is mask code.
-const ErrNoMask C.int = 0xff
 
 // ErrNoExtended is extended errno.
 type ErrNoExtended int
@@ -73,24 +64,6 @@ func (err ErrNo) Error() string {
 // Extend return extended errno.
 func (err ErrNo) Extend(by int) ErrNoExtended {
 	return ErrNoExtended(int(err) | (by << 8))
-}
-
-// Error return error message that is extended code.
-func (err ErrNoExtended) Error() string {
-	return Error{Code: ErrNo(C.int(err) & ErrNoMask), ExtendedCode: err}.Error()
-}
-
-func (err Error) Error() string {
-	var str string
-	if err.err != "" {
-		str = err.err
-	} else {
-		str = C.GoString(C.sqlite3_errstr(C.int(err.Code)))
-	}
-	if err.SystemErrno != 0 {
-		str += ": " + err.SystemErrno.Error()
-	}
-	return str
 }
 
 // result codes from http://www.sqlite.org/c3ref/c_abort_rollback.html
--- a/error_cgo.go
+++ b/error_cgo.go
@@ -1,0 +1,38 @@
+// Copyright (C) 2019 Yasuhiro Matsumoto <mattn.jp@gmail.com>.
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+// +build cgo
+
+package sqlite3
+
+/*
+#ifndef USE_LIBSQLITE3
+#include <sqlite3-binding.h>
+#else
+#include <sqlite3.h>
+#endif
+*/
+import "C"
+
+// ErrNoMask is mask code.
+const ErrNoMask C.int = 0xff
+
+// Error return error message that is extended code.
+func (err ErrNoExtended) Error() string {
+	return Error{Code: ErrNo(C.int(err) & ErrNoMask), ExtendedCode: err}.Error()
+}
+
+func (err Error) Error() string {
+	var str string
+	if err.err != "" {
+		str = err.err
+	} else {
+		str = C.GoString(C.sqlite3_errstr(C.int(err.Code)))
+	}
+	if err.SystemErrno != 0 {
+		str += ": " + err.SystemErrno.Error()
+	}
+	return str
+}
--- a/sqlite3_load_extension_test.go
+++ b/sqlite3_load_extension_test.go
@@ -3,6 +3,7 @@
 // Use of this source code is governed by an MIT-style
 // license that can be found in the LICENSE file.
 
+// +build cgo
 // +build !sqlite_omit_load_extension
 
 package sqlite3
--- a/static_mock.go
+++ b/static_mock.go
@@ -35,3 +35,6 @@ func (c *SQLiteConn) RegisterCommitHook(func() int)                            {
 func (c *SQLiteConn) RegisterFunc(string, interface{}, bool) error             { return errorMsg }
 func (c *SQLiteConn) RegisterRollbackHook(func())                              {}
 func (c *SQLiteConn) RegisterUpdateHook(func(int, string, string, int64))      {}
+
+func (err Error) Error() string         { return "no error message in CGO_ENABLED=0" }
+func (err ErrNoExtended) Error() string { return Error{}.Error() }


rittneje:
@arp242 I'm a little confused by the use case here. If someone isn't compiling with cgo, then presumably they aren't intending to use SQLite. Consequently, your generic `ErrUnique` implementation is compiling in this library for no reason at all, which bloats the binary.

This problem exists independently of the use of cgo. For instance I could have a cgo-enabled binary that is strictly using PostgreSQL. As more dialects of SQL are supported, the resulting binary size will continue to increase, despite the fact that it is all dead code. Thus I think it would make more sense to follow a registry pattern and sidestep the cgo compilation issue altogether.

One possible approach:

error.go
```go
var uniqueErrFuncs []func(err error) bool

func registerUniqueErrFunc(f func(err error) bool) {
    uniqueErrFuncs = append(uniqueErrorFuncs, f)
}

func ErrUnique(err error) bool {
    for _, f := range uniqueErrFuncs {
        if f(err) {
            return true
        }
    }
    return false
}
```

error_sqlite3.go
```go
// +build sqlite3

func init() {
    registerUniqueErrFunc(sqlite3ErrUnique)
}

func sqlite3ErrUnique(err error) bool {
    var sqlErr *sqlite3.Error
    return errors.As(err, &sqlErr) && sqlErr.ExtendedCode == sqlite3.ErrConstraintUnique
}
```

error_postgres.go
```go
// +build postgres

func init() {
    registerUniqueErrFunc(postgresErrUnique)
}

func postgresErrUnique(err error) bool {
    var pqErr *pq.Error
    return errors.As(err, &pqErr) && pqErr.Code == "23505" 
}
```

An alternative to build tags would be to put the dialect-specific code in sub-packages that automatically register themselves, and the client is responsible for underscore importing the necessary sub-packages. (This is how the database/sql package handles it.)

arp242:
Yeah, it's easy to work around with build tags, which is what I've been doing for a while, but I don't think it's especially great; you need to split what is essentially a trivial function out over 2 or 3 files, and the more error codes/SQL engines you add, the more it becomes.

It's a bit of a song and dance for something that should, IMO, Just Work™ without too much fussing about.  The static_mock.go file already contains a few references for this use case, and this seems to fit well to me. It just makes usage a bit easier.

> If someone isn't compiling with cgo, then presumably they aren't intending to use SQLite. Consequently, your generic ErrUnique implementation is compiling in this library for no reason at all, which bloats the binary.

The amount of "bloat" this adds seems negligible – nowhere near where I'm worried about it. The assume the compiler will just compile in one struct with a few methods, which is quite small on the scale of even a small application.

You could also argue that having to use multiple implementations is a kind of "bloat": not in binary size, but in code complexity. Not that it's not very super-duper complicated or anything, but it's certainly a lot harder to follow than just having a function with dead-simple logic.

Besides, if you don't want this bloat in your binary, you can continue to using what you're using before; it's not really preventing people from doing that.

rittneje:
The big downside I see to your change is that it further hides the fact that this library requires cgo until runtime, by "pretending" to compile successfully.

> The amount of "bloat" this adds seems negligible – nowhere near where I'm worried about it. The assume the compiler will just compile in one struct with a few methods, which is quite small on the scale of even a small application.

Like I said, the amount of bloat is going to increase the more dialects of SQL you try to support. It will also bloat as you refer to more types/methods (such as `*sqlite3.Stmt` and `*sqlite3.Conn`), even if the code in question is never really used. I agree that in the specific case of `*sqlite3.Error` and the various error codes this will be minuscule, but it can add up, especially when transitive dependencies are involved.

Can you expand upon your use case? I assume you are intending to write some generic library. Who is calling `ErrUnique` and why is it necessary (as opposed to the client just doing logic against `*sqlite3.Error` themselves)? What other kinds of functionality is your package offering?

arp242:
You often don't know someone wants to use SQLite until runtime anyway; it depends what user tells the application to do (i.e. which database to use), and in many cases it will already compile just fine without cgo – it's just if you happen to use certain parts of the API that it fails. If you really want to prevent people from compiling without cgo then it's not hard to prevent that with a build tag, and that won't complicate the normal logic, and isn't dependent on which parts of the API you may happen to use.

I tested and the "bloat" it adds is 4.2k, or 96 bytes in a stripped binary, when building without cgo and this patch. I really don't understand what the concern is here... "Hello world" in Go is 2M. A "Hello world" importing `pq` and `select 'hello'` is 5.8M. So this adds  what, ~0.0008% or something? This doesn't really strike me as a source of concern.

The use case is that I don't have to juggle around and play games with 3 different `sql.go`, `sql_cgo.go`, and `sql_nocgo.go` to get basic stuff done. Sometimes there is no way around this, such as syscalls, but this isn't really one of those cases. It's a simple fix on go-sqlite3's end which offloads complexity for everyone else.

rittneje:
> You often don't know someone wants to use SQLite until runtime anyway

That depends. If I'm writing top-level code, that is certainly not true - I know my application is specifically using SQLite. Also anything interacting with SQL in the abstract can get tricky because not all dialects of SQL work the same way. (For example, a query written against a SQLite database may or may not even work against a PostgreSQL database, or vice versa.) This is why I'm asking for more information about the use cases that have led you to "I need to know if an error I got from a database driver is a uniqueness violation but I don't know if it's SQLite or PostgreSQL." Are you writing a library? If so, what kinds of functionality is it offering (other than `ErrUnique`)? Why do consumers need to check for uniqueness errors? Why do consumers need a generic dialect-agnostic function for this? (Please note that I'm not trying to dismiss your use case or proposed resolution as invalid. I'm just trying to fully understand the problem you are facing.)

> I tested and the "bloat" it adds is 4.2k, or 96 bytes in a stripped binary, when building without cgo and this patch. I really don't understand what the concern is here...

Yes, I mentioned above that I expected it to be minuscule in this specific case. But it doesn't really address what I see as the larger problem. For instance, anyone who uses `ErrUnique` with cgo _enabled_ has now compiled the entirely of SQLite (the C library) into their application, even if they are only using PostgreSQL. On top of that, if you are using Go modules, now consumers of the library have to download all the driver code even though it's not really being used. This isn't really a big problem when you only have SQLite and PostgreSQL, but you can imagine that it gets worse as you support more drivers for other dialects.

arp242:
I just don't like having three files for a simple function. In any application. In any circumstance. I think it's not needed and that these sort of things are better handled here, rather than in everyone's applications. SQL portability issues and whatnot are unrelated.

> For instance, anyone who uses ErrUnique with cgo enabled has now compiled the entirely of SQLite (the C library) into their application.

If you care about having special builds to shave off something on your filesize then this patch changes nothing about that (other than the 4.2k/96b mentioned before).

> On top of that, if you are using Go modules, now consumers of the library have to download all the driver code even though it's not really being used.

This is already the case? I don't follow how this patch changes anything about that?

rittneje:
> This is already the case? I don't follow how this patch changes anything about that?

I wasn't referring to your change set. I'm saying that your current solution of having `ErrUnique` statically refer to both drivers has the negative consequences described. That solution is also what leads you to needing a non-cgo version of `*sqlite3.Error`.

> I just don't like having three files for a simple function. In any application. In any circumstance.

Suppose there is some syscall that is a one liner for Linux and Windows. This would require separate files because neither can compile for the other system. It seems that your annoyance at having three files is more because there isn't an `#ifdef` preprocessor macro.

> I think it's not needed and that these sort of things are better handled here, rather than in everyone's applications.

Who is "everyone"? It seems to me the only time this issue comes up is when someone attempts to write some dialect-agnostic library, which inevitably causes an issue due to cgo. But to me this is an example of an XY problem, because the real issue would seem to be the design flaw of trying to compile all the drivers in question, even though most aren't being used. The cgo failure is just a symptom.

> SQL portability issues and whatnot are unrelated.

The reason I brought this up is because the only time `ErrUnique` even accomplishes something seems to be when I am running queries against a database but I don't know if the dialect is SQLite or PostgreSQL. But that use case seems contrived because in order to write those queries you already know the dialect. So what's the point of having `ErrUnique` at all? In other words, what is the use case in which you have dialect-agnostic queries and need to know if they are reporting uniqueness constraint violations?

> If you care about having special builds to shave off something on your filesize then this patch changes nothing about that (other than the 4.2k/96b mentioned before).

I think you misunderstood what I was saying. The 4.2k number you calculated is the difference between not compiling against this library at all, and compiling against the non-cgo version you have in your changeset. I am referring to the bloat of the cgo _enabled_ version, which seems to be over 3MB. (Imagine an application using your library that only uses PostgreSQL, and also uses cgo for some other purpose.) In other words, the desire to not compile against this library has more use cases than just "cgo disabled".

Of course, one way to deal with this would be to have a special build tag like `sqlite3_stub` in this library that implements the same API (or a subset of it) with "fake" functions and suppresses the cgo version. But to me, that seems confusing. It would be like the golang.org/x/sys/unix package implementing stubs for Windows.

arp242:
I find this entire conversation rather weird to be honest; I just have a program that can connect to multiple SQL engines, which seems like a common and normal use case and this makes that a wee bit easier. "Compile with cgo but not SQLite" is not a feature I need or care about.

If you want to solve things in a different way or do care about it: great! Go for it. This patch isn't preventing you from doing so.

rittneje:
That isn't really a common use case. Most people tend to write applications that use a specific SQL dialect, because attempting to write queries that are dialect agnostic is hard and not usually worth the effort. And for libraries that do have to manage multiple dialects and thus multiple drivers, they handle this with either build flags or separate packages, like gorm does.

As I mentioned before, there are a few problems with your patch:
1. This library cannot really compile without cgo. Your patch obfuscates this, pushing mistakes from compile-time detection to run-time detection. Note that cgo is automatically disabled when cross-compiling, so this has great potential to cause confusion.
2. I still don't fully understand your use case. From your latest post, you say you have an application that needs to be able to connect to both SQLite databases and PostgreSQL databases, but then why aren't you compiling with cgo all the time?
3. If you are intending to write some generic reusable library that can be used with both SQLite databases and PostgreSQL databases, then the decision to import this library or not is distinct from the decision to use cgo. To reiterate, suppose my application only uses PostgreSQL, and also needs cgo for some unrelated purpose. In order to use your library, I am now forced to bloat my application by 3MB. This is a flaw in your library. If your library provided a mechanism to not import this library in such a circumstance, then trivially the cgo issue would also be solved.

I understand that you aren't really interested in addressing (3) at this time, and that this patch doesn't prevent someone else from addressing (3). That's fine. But your patch is still a reflection of a design choice in your code, so I think the logic should remain there. And if you think about it, your patch is kind of hacky anyway. In reality, you don't expect `*sqlite3.Error` to ever show up in an application that isn't using cgo, because (a) you can't use this driver without cgo, and (b) the type isn't real, as evidenced by the awkward way you had to implement the `Error` method.

arp242:
> In order to use your library, I am now forced to bloat my application by 3MB.

Don't use it then... You're not "forced" to do anything.

Thousands upon thousands of applications can run against multiple SQL engines; while others are tied so a specific one. There are pros and cons to both, but it's hardly an uncommon thing to do.

rittneje:
It doesn't really matter to me what you choose to do in your own application or library. The original reason you made this changeset was because you wanted your code to still compile and have certain behavior in the absence of cgo. If you feel that decision makes the most sense in your case, that totally your prerogative. But making the change against this library has implications for _everyone_.

In my experience, the majority of use cases for this library are people who are exclusively using SQLite, so not using cgo truly is an error. And since cgo gets disabled by default while cross-compiling, it would seem undesirable to change the existing behavior (pushing failures from compile-time to run-time).

I understand there are also applications such as yours in which there are multiple drivers in play. As far as I am aware, most of these avoid compiling in unnecessary drivers in the first place, so they don't suffer from any cgo problems directly. (If you specifically try to reference SQLite but don't enable cgo, then you end up with an error, which is probably desirable in that case.) You have chosen not to handle that generic situation, and instead have focused specifically on the cgo problem. Again, that's entirely up to you. But since you are deciding what it means to compile the overall application without cgo, it seems reasonable to me that you are responsible for maintaining the cgo vs. non-cgo split. I understand that having to maintain two separate really short files is slightly annoying, but that is the only option Go affords, regardless of where that split lives.

In short, this change has negative consequences for what I believe to be the most common use case of this library, and is of no value for those libraries that are designed to handle multiple drivers optimally. Do you have examples of other applications/libraries like yours, where they always want to compile against multiple drivers and only have a problem with cgo?

arp242:
"Compiles without cgo" is already a feature of the library; this merely expands on that. This is why static_mock.go has:

    // +build !cgo
    func init() {
        sql.Register("sqlite3", &SQLiteDriver{})
    }

If you're going to provide a shim like that then you might as well make a complete one, instead of having half the library work without cgo via shims while the other half doesn't – correcting that (or at least, part of it) is what this patch does

If you want to have a "I want a compiletime error if people use CGO_ENABLED=0" feature; sure, that would be useful; I actually have a a [little hack](https://stackoverflow.com/a/65137236/660921) to make sure people don't use <Go 1.13 to prevent runtime errors relating to the new error handling (Debian ships with Go 1.11, had some people be very confused) and I'm all for compiletime errors vs. runtime-errors when possible, but making other usage patterns harder isn't the right way to go about that IMO, especially not by stopping the expansion of an existing feature. What you want is essentially a new feature.

We could, for example, move the registry of that mock driver to a new `github.com/mattn/go-sqlite3/shim` subpackage or some such, and leave the shim types/functions as-is (although that would be a breaking change). Or if you want "works without cgo" by default create a new subpackage you can import which triggers a compile-time warning (which would be a compatible change).

But again, all of that is a new feature.

supermarin:
@arp242 [pretty unrelated] stumbled upon this issue since I ended up using the same approach. Just wanted to point out that `	var sqlErr *sqlite3.Error` won't ever work because `errors.As(err, &sqlErr)` already passes the address of `sqlErr` and tries to unpack a pointer instead of `sqlite3.Error`.  Using just `var sqlErr sqlite3.Error` works as intended

PR #901, opened 2021-01-03
Export sqlite3_serialize & sqlite3_deserialize

otoolep:


otoolep:
This code passes testing fine on my local machine (Ubuntu 16), so I'm confused about the failures here. Any help? For example:
`./sqlite3.go:943: could not determine kind of name for C.sqlite3_deserialize`

rittneje:
@otoolep The builds are failing when dynamically linked against the system-provided SQLite library (as opposed to statically linked against the version provided inside this wrapper library). `sqlite3_serialize` and `sqlite3_deserialize` were not added until v3.23.0, and even then are only defined if you compile with `SQLITE_ENABLE_DESERIALIZE`. You will need to account for this.

otoolep:
Thanks @rittneje -- I already added SQLITE_ENABLE_DESERIALIZE.

How do you "account for it"? Have you an example I could follow?

rittneje:
You only added `SQLITE_ENABLE_DESERIALIZE` when statically compiling against the version included in this library. Unfortunately, if the user opts for dynamic linking, you have no control over which version is being used, or whether `SQLITE_ENABLE_DESERIALIZE` was defined when the shared library was originally compiled. I think you'll have to use `#if` on a combination of the `SQLITE_VERSION_NUMBER` macro and the `SQLITE_ENABLE_DESERIALIZE`. You will also have to figure our what your two new functions do if the corresponding C functions don't exist.

Something like:
```C
#if SQLITE_VERSION_NUMBER < 3023000 || !defined (SQLITE_ENABLE_DESERIALIZE)
unsigned char *sqlite3_serialize(...) {
    // return null?
    ...
}

int sqlite3_deserialize(...) {
    // return error code?
    ...
}
#endif
```

otoolep:
I see, thank you @rittneje

otoolep:
I'll revisit this, as you suggest. Closing for now.

PR #908, opened 2021-01-11
Adds a fuzz target

catenacyber:
Cf https://github.com/google/oss-fuzz/pull/4946

This will allow oss-fuzz integration, and thus continuous fuzzing.

@mattn does this fuzz target look good to you ?

### DIFF
--- a/_example/fuzz/fuzz_openexec.go
+++ b/_example/fuzz/fuzz_openexec.go
@@ -1,0 +1,30 @@
+package sqlite3_fuzz
+
+import (
+	"bytes"
+	"database/sql"
+	"io/ioutil"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+func FuzzOpenExec(data []byte) int {
+	sep := bytes.IndexByte(data, 0)
+	if sep <= 0 {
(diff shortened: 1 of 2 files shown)


catenacyber:
If we only call `C.sqlite3_close_v2(db)` and not `conn.Close()`, we do not call `deleteHandles` which frees some memory with `C.free(handle)`

I guess other lines have to be fixed this way

catenacyber:
Friendly ping on this PR @mattn  ?

mattn:
Thank you

mattn:
Hmm, go-acc refer older version of hashicorp/hcl. So CI got fails. @catenacyber Can you handle this issue?

catenacyber:
I do not see the CI failure for this PR (which also does not mention neither go-acc nor hashicorp)...
So, I am not sure what you mean... Could you provide me some pointers ?

mattn:
All tests are failing https://github.com/mattn/go-sqlite3/runs/1903852468

catenacyber:
I created https://github.com/mattn/go-sqlite3/pull/915

PR #909, opened 2021-01-22
Update SQLite3_ The columntypescantype method of type

Auler:
Update SQLite3_ The columntypescantype method of type,
The original method always returns nil

### DIFF
--- a/sqlite3_type.go
+++ b/sqlite3_type.go
@@ -1,8 +1,3 @@
-// Copyright (C) 2019 Yasuhiro Matsumoto <mattn.jp@gmail.com>.
-//
-// Use of this source code is governed by an MIT-style
-// license that can be found in the LICENSE file.
-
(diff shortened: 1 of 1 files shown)


sqlite3_type.go
```diff
@@ -1,8 +1,3 @@
-// Copyright (C) 2019 Yasuhiro Matsumoto <mattn.jp@gmail.com>.
-//
-// Use of this source code is governed by an MIT-style
-// license that can be found in the LICENSE file.
-
```
mattn, line 5: Please do  not delete copyright.

Auler: Sorry, I didn't notice that.

Auler, line 5: add copyright

Auler:
Restore the mistakenly deleted copyright

PR #915, opened 2021-02-16
Go get go-acc with environment variable for go modules

catenacyber:
Cf https://github.com/mattn/go-sqlite3/pull/908#issuecomment-779500688

This does not seem related to the addition of the fuzz target...


mattn:
Thank you

PR #919, opened 2021-02-21
Adds CIFuzz for fuzzing as continuous integration

catenacyber:
This will allow to catch bugs found by short fuzzing in Pull Requests before they get merged

For instance : https://github.com/protocolbuffers/upb/pull/369


catenacyber:
oss-fuzz go projects compilation is broken at the moment but should be fixed once https://github.com/google/oss-fuzz/pull/5228 gets to the right Docker images

Issue #920, opened 2021-02-25
need to error handler on example


huf0813:
on example/simple directory, i found a possibility to improve on error handler. Can i just add defer func to close function? Because it will return some error that need to be handled. Thx

PR #921, opened 2021-02-25
add error handler on defer and func with error return

huf0813:
This PR is related to issues #920. Please kindly check the code. Thx and have a good one.

### DIFF
--- a/_example/simple/simple.go
+++ b/_example/simple/simple.go
@@ -9,13 +9,19 @@ import (
 )
 
 func main() {
-	os.Remove("./foo.db")
+	if err := os.Remove("./foo.db"); err != nil {
+		log.Fatal(err)
+	}
 
 	db, err := sql.Open("sqlite3", "./foo.db")
 	if err != nil {
 		log.Fatal(err)
 	}
-	defer db.Close()
+	defer func() {
+		if err := db.Close(); err != nil {
+			log.Fatal(err)
+		}
+	}()
 
 	sqlStmt := `
 	create table foo (id integer not null primary key, name text);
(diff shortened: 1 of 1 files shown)


_example/simple/simple.go
```diff
@@ -9,13 +9,19 @@ import (
 )
 
 func main() {
-	os.Remove("./foo.db")
+	if err := os.Remove("./foo.db"); err != nil {
```
rittneje, line 12: Need to do `err != nil && !os.IsNotExist(err)`.

_example/simple/simple.go
```diff
@@ -35,20 +41,30 @@ func main() {
...
+		}
+	}()
 	for i := 0; i < 100; i++ {
 		_, err = stmt.Exec(i, fmt.Sprintf("こんにちわ世界%03d", i))
 		if err != nil {
 			log.Fatal(err)
 		}
 	}
-	tx.Commit()
+	if err := tx.Commit(); err != nil {
```
rittneje, line 55: You should defer `tx.Rollback()` right after `db.Begin()` above for completeness. Only log if it returns an error other than `sql.ErrTxDone`.

_example/simple/simple.go
```diff
@@ -35,20 +41,36 @@ func main() {
...
 	for i := 0; i < 100; i++ {
 		_, err = stmt.Exec(i, fmt.Sprintf("こんにちわ世界%03d", i))
 		if err != nil {
 			log.Fatal(err)
 		}
 	}
-	tx.Commit()
+	errCommit := tx.Commit()
+	defer func() {
+		if err := tx.Rollback(); err != nil {
```
rittneje, line 57: This should be in a `defer` right after `db.Begin()` (and its err check), since we should do it if we fail at any point during the transaction.

_example/simple/simple.go
```diff
@@ -35,20 +41,36 @@ func main() {
...
+			log.Fatal(err)
+		}
+	}()
+	if errCommit != nil && errCommit != sql.ErrTxDone {
```
rittneje, line 61: Here `sql.ErrTxDone` would be a legitimate error because it should not be possible. (The deferred `tx.Rollback()` should not consider it fatal though since we truly could have already committed.)

PR #929, opened 2021-03-21
Update almagation 3.36 and correct mac OS problem compilation

realPy:
Mac os compilation work with --tags="darwin libsqlite3" but not with --tags=darwin because assert is not defined with clang.   
This  PR correct this bug and now libsqlite3 can be included static (not only link with the libsqlite3).  
I have also update the almagation to the latest 3.35.2 version

### DIFF
--- a/sqlite3-binding.h
+++ b/sqlite3-binding.h
@@ -124,9 +124,9 @@ extern "C" {
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.35.4"
-#define SQLITE_VERSION_NUMBER 3035004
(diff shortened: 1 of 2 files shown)


mattn:
Thank you. I hope to update alamagation code my self. I'll merge your few commits in this PR in later.

realPy:
Update to almagation 3.36

mattn:
Addressed in https://github.com/mattn/go-sqlite3/pull/955

Why did you squashed or removed some commits for macOS?

realPy:
Because some of you tag and commit was missing in my fork.
I just get a clean copy of your work and apply a patch

realPy:
to correct macos native compilation (embbed in executable) 
the 3 lines next is need in sqlite3-binding.c (at head)

#ifdef __clang__
#define assert(condition) ((void)0)
#endif

mattn:
Could you please try https://github.com/mattn/go-sqlite3/pull/955 ? #955 include your patch to generator of amalgamation codes.

realPy:
Yes :) i will on

realPy:
It's ok for me , all my test unit passed sucessfull passed with #955

PR #941, opened 2021-04-18
Resolve windows CI issues

aeneasr:
Do not use `-u` flag when fetching go-acc


aeneasr:
Windows still fails but not for reasons related to go-acc 😓

PR #946, opened 2021-05-05
Add error checking in simple example for tx.Commit

lnksz:
Based on https://golang.org/pkg/database/sql/#Tx.Commit this function returns an error type.
So why not check it.


mattn:
Thank you

PR #960, opened 2021-07-24
add PreConnectExtensions, PreConnectHook

satotake:
Some extensions such as VFSes need to be loaded before opening db file.
In the other languages' sqlite3 binding, this can be done by

1. opening temporary db
2. loading extension
3. connecting actual db file

, though this seems to be tricky.

I have checked in [rust](https://github.com/rusqlite/rusqlite) and [python](https://docs.python.org/3/library/sqlite3.html).
For example in rust, we can write it like the below

```rust
let conn_empty = Connection::open_in_memory()?;
conn_empty.load_extension("./vfsext", Some("sqlite3_vfsext_init"));
let conn = Connection::open(TARGET)?;
```

Unfortunately, this approach does not work with `go-sqlite3`.
I am not sure but this may be because of requirements as driver of `database/sql`.

Thus, `PreConnectExtensions` and `PreConnectHook` are added in the similar to rust example.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -290,8 +290,10 @@ const (
 
 // SQLiteDriver implements driver.Driver.
 type SQLiteDriver struct {
-	Extensions  []string
-	ConnectHook func(*SQLiteConn) error
+	Extensions           []string
+	ConnectHook          func(*SQLiteConn) error
+	PreConnectExtensions []string
+	PreConnectHook       func(*SQLiteConn) error
 }
 
 // SQLiteConn implements driver.Conn.
@@ -1385,7 +1387,62 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 		}
 	}
 
-	var db *C.sqlite3
+	open := func(name *C.char, vfs *C.char) (*C.sqlite3, error) {
+		var db *C.sqlite3
+		rv := C._sqlite3_open_v2(name, &db,
+			mutex|C.SQLITE_OPEN_FULLMUTEX|C.SQLITE_OPEN_READWRITE|C.SQLITE_OPEN_CREATE,
+			vfs)
+
+		if rv != 0 {
+			// Save off the error _before_ closing the database.
+			// This is safe even if db is nil.
+			err := lastError(db)
+			if db != nil {
+				C.sqlite3_close_v2(db)
+			}
+			return nil, err
+		}
+
+		if db == nil {
+			return nil, errors.New("sqlite succeeded without returning a database")
+		}
+
+		rv = C.sqlite3_busy_timeout(db, C.int(busyTimeout))
+		if rv != C.SQLITE_OK {
+			C.sqlite3_close_v2(db)
+			return nil, Error{Code: ErrNo(rv)}
+		}
+		return db, nil
+	}
+
+	if len(d.PreConnectExtensions) > 0 || d.PreConnectHook != nil {
+		// create temp memory db for loading extensions before opening
+		name_temp := C.CString(":memory:")
+		defer C.free(unsafe.Pointer(name_temp))
+
+		db_temp, err := open(name_temp, nil)
+		defer C.sqlite3_close_v2(db_temp)
+		if err != nil {
+			return nil, err
+		}
+
+		conn_temp := &SQLiteConn{db: db_temp}
+		defer conn_temp.Close()
+
+		if len(d.PreConnectExtensions) > 0 {
+			if err := conn_temp.loadExtensions(d.PreConnectExtensions); err != nil {
+				return nil, err
+			}
+		}
+
+		if d.PreConnectHook != nil {
+			if err := d.PreConnectHook(conn_temp); err != nil {
+				conn_temp.Close()
+				return nil, err
+			}
+		}
+	}
+
 	name := C.CString(dsn)
 	defer C.free(unsafe.Pointer(name))
 	var vfs *C.char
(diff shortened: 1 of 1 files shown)


rittneje:
I don't understand why this would be a hook. If you need to do something before connecting to the database, you can just do it prior to calling `sql.Open` or `sql.OpenDB`. I wouldn't think you'd want to re-register a VFS every time a new connection is going to be opened.

Note that the existing `ConnectHook` is needed to configure the new connection each time a new one is established, since database/sql is really modeling a connection pool. For example, prior to the introduction of the `_fk` DSN option, the only way to ensure foreign keys got enabled on all connections was to do so via a hook.

satotake:
thanks @rittneje 

> Note that the existing ConnectHook is needed to configure the new connection each time a new one is established, since database/sql is really modeling a connection pool. For example, prior to the introduction of the _fk DSN option, the only way to ensure foreign keys got enabled on all connections was to do so via a hook.

I see.

I added `PreConnectHook` for users who would like to use the optional second argument of `loadExtension` (entry name), 
referring [this example](https://github.com/dinedal/go-sqlite3-extension-functions/blob/master/go/extension-functions.go).

In most cases, we have chance to edit entry name of extension in c file directly.
So I do not hesitate to remove `PreConnecHook` .

rittneje:
@satotake If I understand correctly, these are the problems you are trying to address:

1. You have a custom extension that itself registers a custom VFS. Because of how SQLite itself (the C library) works, you basically need a dummy connection to load the extension to register the VFS, so you can then use that custom VFS for your real connection.
2. `SQLiteDriver.Extensions` is just a list of strings, so there's no way for you to specify a custom entrypoint to `sqlite3_load_extension`.

This "dummy connection" logic appears to be a standard mechanism for registering custom VFS. For example, this very process is described here: https://www.sqlite.org/cksumvfs.html In that case would it make more sense to add a specific function to this library to allow you to load an extension exclusively for its side effects? Internally it could just make some dummy connection but the client need not care about this. This also prevents your extension from getting reloaded/re-registered whenever a new connection is established via database/sql.

satotake:
@rittneje 

> If I understand correctly, these are the problems you are trying to address:

Your understanding is right.

> This "dummy connection" logic appears to be a standard mechanism for registering custom VFS. 

I was trying to find any other better ways for a long time. Thanks.

> In that case would it make more sense to add a specific function to this library to allow you to load an extension exclusively for its side effects?

If it is possible, it will be neat.
I suspect that this might be impossible because `database/sql` may prepares separated processes for some reasons (thread safety?) and we, in golang and sqlite, have to create fresh dummy db every time when `sql.Open`.


Anyway, I am trying what you said.

rittneje:
> I suspect that this might be impossible because database/sql may prepares separated processes for some reasons (thread safety?) and we, in golang and sqlite, have to create fresh dummy db every time when sql.Open .

database/sql does not make separate processes. It can establish multiple connections, but they will all be from your single process. That is why having the driver treat this as a normal connection hook would be a problem.

satotake:
Walking through `sql/database` code, I may be understanding what you said finally. 
You are suggesting to implement `Connector` interface and etc, right?

rittneje:
No. Everything in the database/sql/driver package pertains to opening a single connection within the pool. You are trying to do something once _globally_ (not once per connection), so the abstractions provided by database/sql will not help you.

satotake:
hmm, confused again.

> (not once per connection)

It is ideal that we can configure extensions per connection. But how to do it before opening connection?
I admit that `Connector` is more global than connection but less global than `sql.Register`.

rittneje:
> It is ideal that we can configure extensions per connection.

You aren't trying to configure extensions per connection. When you register a VFS, it gets registered globally, not per connection. See https://www.sqlite.org/c3ref/vfs_find.html.

Under normal circumstances, you would just call `sqlite3_vfs_register`. However, instead this VFS has chosen to bundle itself as an extension, which internally calls `sqlite3_vfs_register` for you. But SQLite requires some connection in to call [`sqlite3_load_extension`](https://www.sqlite.org/c3ref/load_extension.html). This connection is totally irrelevant to your VFS extensions, because it is just using it as a means to an end. That is why the docs say to create a dummy in-memory database connection that gets thrown away.

So, you need to first load your extension once per run of your application, solely for the side effect of it calling `sqlite3_vfs_register`. Once that happens, you can use that VFS for your actual connections. You can still open other connections that choose not to use that VFS.

To do this, you don't need to use database/sql at all. For example, it might look like this:

```go
package sqlite3

func LoadExtensionForSideEffects(string path) error {
    c, err := &Driver{Extensions: []string{path}}.Open(":memory:")
    if err != nil {
        return err
    }
    c.Close()
    return nil
}
```
Note: That function name is just for demonstration purposes.

To be clear, this hypothetical new function is just for client convenience. Nothing stops you from writing similar code in your own main.go right now (the extension entrypoint name issue aside).

satotake:
It works !
I can not understand why go code similar to the previous rust example does not have any effect while this is valid side effect.
anyway thanks. I learned a lot from you.

rittneje:
> I can not understand why go code similar to the previous rust example does not have any effect

Can you post the Go code that you were trying with that didn't work?

satotake:
Of cource

```golang
package main

import (
	"database/sql"
	"fmt"
	"log"
	"net/url"

	"github.com/mattn/go-sqlite3"
)

func init() {
	sql.Register("sqlite3-ext", &sqlite3.SQLiteDriver{
		Extensions: []string{"./extension"},
	})
}

func main() {
	sql.Open("sqlite3-ext", "memory")

	uri := "some uri"
	db, err := sql.Open("sqlite3-ext", uri)
	if err != nil {
		log.Fatal(err)
	}

	rows, err := db.Query("select tbl_name from sqlite_master;")
	if err != nil {
		log.Fatal(err)
	}

	for rows.Next() {
		var name string
		err = rows.Scan(&name)
		if err != nil {
			log.Fatal(err)
		}
		fmt.Println(name)

	}
	err = rows.Close()
	if err != nil {
		log.Fatal(err)
	}
}

```

rittneje:
Ah, the reason that didn't work is because database/sql is lazy and doesn't open a connection until it has to. If you change it to this, then it should work.
```go
dummy, err := sql.Open("sqlite3-ext", "memory")
if err != nil {
    ...
}
// Ping forces database/sql to actually open a driver connection.
if err  := dummy.Ping(); err != nil {
    ...
}
dummy.Close()

// This doesn't need to use your "sqlite3-ext" driver
// since the dummy connection has already registered
// the custom VFS.
db, err := sql.Open("sqlite3", uri)
...
```

satotake:
> Ah, the reason that didn't work is because database/sql is lazy 

Totally clear!
This was my misunderstanding...
Current [examples](https://github.com/mattn/go-sqlite3/blob/master/_example/mod_regexp/extension.go) is misleading somehow in my opnion.
This registers it as another name but this does not matter because this affects globally (if my understanding is correct)

> To be clear, this hypothetical new function is just for client convenience. Nothing stops you from writing similar code in your own main.go right now (the extension entrypoint name issue aside).

I think it is better to add the hypothetical functions as top-level one to clarify what happens here and improve the example

It seems that some developers have similar misunderstanding to mine.

satotake:
Edit: previous comment

It depends on extensions that it has global effect or not. So this example is right and I was wrong again.
I appreciate @rittneje 's effort here.

PR #989, opened 2021-12-16
Update to SQLite 3.37.0.

dsymonds:
This adds support for STRICT tables (https://sqlite.org/stricttables.html).


mattn:
Thanks. Please let me do it in my-self.

mattn:
I wonder why my PR is broken.

https://github.com/mattn/go-sqlite3/pull/991

Issue #999, opened 2022-01-15
Support sqlite3_file_control() 


benbjohnson:
Hi @mattn, thanks for the great library. I use it all the time.

I'm using `go-sqlite3` in my [Litestream](https://github.com/benbjohnson/litestream) project for SQLite replication and I've been doing a hack to persist the WAL file across restarts where I just don't call `Close()` on the database. Definitely hacky but it works. However, I recently learned there's `sqlite3_file_control()` opcode called [`SQLITE_FCNTL_PERSIST_WAL`](https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlpersistwal) that does the same thing.

Is there an existing way to execute `sqlite3_file_control()` in `go-sqlite3`? I didn't see one but I thought I'd ask.

I'm happy to add the functionality. Are you open to accepting a patch for it?

rittneje:
Seems fine. But do be mindful that some users compile against their system installation of SQLite, meaning that not all constants/functions from the statically compiled header file are guaranteed to exist in all cases.

benbjohnson:
@rittneje I pushed up a [pull request](https://github.com/mattn/go-sqlite3/pull/1000) to add a `FileControl()` method to `SQLiteConn` and a corresponding test. I also tested it against Litestream in this [`persist-wal`](https://github.com/benbjohnson/litestream/compare/persist-wal) branch. Let me know if there's anything you'd like me to change with the PR. Thanks!

benbjohnson:
@rittneje @mattn Thanks so much for all the reviews! 🙏

PR #1000, opened 2022-01-16
Add sqlite3_file_control() support

benbjohnson:
This pull request adds the `SQLiteConn.FileControl()` method which calls the underlying `sqlite3_file_control()` function. This can be used for low-level operations on SQLite databases such as persisting the WAL file after database close.

Fixes https://github.com/mattn/go-sqlite3/issues/999

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -297,6 +297,51 @@ const (
 	/*SQLITE_RECURSIVE           = C.SQLITE_RECURSIVE*/
 )
 
+// Standard File Control Opcodes
+// See: https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html
+const (
+	SQLITE_FCNTL_LOCKSTATE             = int(1)
+	SQLITE_FCNTL_GET_LOCKPROXYFILE     = int(2)
+	SQLITE_FCNTL_SET_LOCKPROXYFILE     = int(3)
+	SQLITE_FCNTL_LAST_ERRNO            = int(4)
+	SQLITE_FCNTL_SIZE_HINT             = int(5)
+	SQLITE_FCNTL_CHUNK_SIZE            = int(6)
+	SQLITE_FCNTL_FILE_POINTER          = int(7)
+	SQLITE_FCNTL_SYNC_OMITTED          = int(8)
+	SQLITE_FCNTL_WIN32_AV_RETRY        = int(9)
+	SQLITE_FCNTL_PERSIST_WAL           = int(10)
+	SQLITE_FCNTL_OVERWRITE             = int(11)
+	SQLITE_FCNTL_VFSNAME               = int(12)
+	SQLITE_FCNTL_POWERSAFE_OVERWRITE   = int(13)
+	SQLITE_FCNTL_PRAGMA                = int(14)
+	SQLITE_FCNTL_BUSYHANDLER           = int(15)
+	SQLITE_FCNTL_TEMPFILENAME          = int(16)
+	SQLITE_FCNTL_MMAP_SIZE             = int(18)
+	SQLITE_FCNTL_TRACE                 = int(19)
+	SQLITE_FCNTL_HAS_MOVED             = int(20)
+	SQLITE_FCNTL_SYNC                  = int(21)
+	SQLITE_FCNTL_COMMIT_PHASETWO       = int(22)
+	SQLITE_FCNTL_WIN32_SET_HANDLE      = int(23)
+	SQLITE_FCNTL_WAL_BLOCK             = int(24)
+	SQLITE_FCNTL_ZIPVFS                = int(25)
+	SQLITE_FCNTL_RBU                   = int(26)
+	SQLITE_FCNTL_VFS_POINTER           = int(27)
+	SQLITE_FCNTL_JOURNAL_POINTER       = int(28)
+	SQLITE_FCNTL_WIN32_GET_HANDLE      = int(29)
+	SQLITE_FCNTL_PDB                   = int(30)
+	SQLITE_FCNTL_BEGIN_ATOMIC_WRITE    = int(31)
+	SQLITE_FCNTL_COMMIT_ATOMIC_WRITE   = int(32)
+	SQLITE_FCNTL_ROLLBACK_ATOMIC_WRITE = int(33)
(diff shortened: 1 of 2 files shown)


sqlite3.go
```diff
@@ -1813,6 +1858,23 @@ func (c *SQLiteConn) SetLimit(id int, newVal int) int {
 	return int(C._sqlite3_limit(c.db, C.int(id), C.int(newVal)))
 }
 
+// FileControl invokes the xFileControl method on a given database. The dbName
+// is the name of the database. It will default to "main" if left blank. The op
+// is one of the opcodes prefixed by "SQLITE_FCNTL_". The arg argument and
+// return code are both opcode-specific. Please see the SQLite documentation.
+//
+// See: sqlite3_file_control, https://www.sqlite.org/c3ref/file_control.html
+func (c *SQLiteConn) FileControl(dbName string, op int, arg unsafe.Pointer) int {
```
rittneje, line 1867: The last argument seems somewhat troubling to me. For example, for some of the ops, it needs to be a pointer to some C struct, but the caller will be hard-pressed to do that within the confines of Go. Consequently, I'm not sure we really want to expose `sqlite3_file_control` directly like this.

Taking a page from the syscall package, we could do something like `FileControlInt` or `FileControlInt`, and use a more user-friendly type for the final parameter, performing the conversion to pointer internally.

Alternatively, we can make a method specifically for getting/setting the persistent WAL mode.

sqlite3_test.go
```diff
@@ -1722,6 +1723,44 @@ func TestAuthorizer(t *testing.T) {
...
+func TestFileControl(t *testing.T) {
+	t.Run("PERSIST_WAL", func(t *testing.T) {
+		tempFilename := TempFilename(t)
+		defer os.Remove(tempFilename)
+
+		sql.Register("sqlite3_FCNTL_PERSIST_WAL", &SQLiteDriver{
+			ConnectHook: func(conn *SQLiteConn) error {
+				arg := int(1)
+				if rc := conn.FileControl("", SQLITE_FCNTL_PERSIST_WAL, unsafe.Pointer(&arg)); rc != 0 {
+					t.Fatalf("Unexpected return code from FileControl(): %d", rc)
```
rittneje, line 1735: You probably should not call `t.Fatalf` here, since this maybe happening on some other goroutine, depending how database/sql works. Instead, just return an error. Below, call `db.Ping` after `sql.Open` and move the assertion there.

sqlite3.go
benbjohnson: Good call. I switched it to `FileControlInt()` instead. 👍

benbjohnson:
@rittneje I fixed up the PR. Can you take another look please?

sqlite3.go
```diff
@@ -297,6 +297,51 @@ const (
 	/*SQLITE_RECURSIVE           = C.SQLITE_RECURSIVE*/
 )
 
+// Standard File Control Opcodes
+// See: https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html
+const (
+	SQLITE_FCNTL_LOCKSTATE             = 1
```
mattn, line 303:
```
SQLITE_FCNTL_LOCKSTATE = int(C.SQLITE_FCNTL_LOCKSTATE)
```

benbjohnson: Fixed 👍

sqlite3.go
```diff
@@ -1813,6 +1858,24 @@ func (c *SQLiteConn) SetLimit(id int, newVal int) int {
 	return int(C._sqlite3_limit(c.db, C.int(id), C.int(newVal)))
 }
 
+// FileControl invokes the xFileControl method on a given database. The dbName
+// is the name of the database. It will default to "main" if left blank. The op
+// is one of the opcodes prefixed by "SQLITE_FCNTL_". The arg argument and
+// return code are both opcode-specific. Please see the SQLite documentation.
+//
+// See: sqlite3_file_control, https://www.sqlite.org/c3ref/file_control.html
+func (c *SQLiteConn) FileControlInt(dbName string, op int, arg int) int {
```
rittneje, line 1867: I think we should name this `SetFileControlInt` or similar, since you are only allowing for them to set the value. (For example, to get the current persistent WAL mode, you need to pass a pointer to -1, and then SQLite will change it to point to the current mode, but as written it won't actually return this value.)
Also, it should return `error`, not `int`. See `lastError`. On that note, you should probably document that this isn't thread-safe, or you need to take the lock to make it so. (It's really `c.lastError` that is not thread-safe, in the sense that that if some other operation happens first, it will yield the wrong result.)

benbjohnson: @rittneje Thanks for the feedback. I updated the method name, changed the return type to `error`, and noted in the docs that the method is not thread-safe.

benbjohnson:
@rittneje @mattn Are there any other changes I should make to get this merged? Thanks!

sqlite3.go
benbjohnson: @mattn Good call. I updated it and push up a new commit. :+1

mattn:
@rittneje good to merge?

sqlite3.go
```diff
@@ -297,6 +297,47 @@ const (
 	/*SQLITE_RECURSIVE           = C.SQLITE_RECURSIVE*/
 )
 
+// Standard File Control Opcodes
+// See: https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html
+const (
```
rittneje, line 302: @mattn By referencing the actual C constants here, this won't compile for anyone linking with an older version of SQLite (via the libsqlite3 tag). For example, `SQLITE_FCNTL_SIZE_LIMIT` wasn't added until 3.27.0.

How do you want to handle this?

mattn: Ah, I didn't consider it. @benbjohnson Sorry much. Could you please rollback changes for the constants?

benbjohnson: @mattn @rittneje No problem. I reverted the code back to using hardcoded integers. 👍

rittneje (approved):
LGTM

mattn:
Thank you

benbjohnson:
@mattn @rittneje Thanks so much for all the reviews! 🙏

PR #998, opened 2022-01-14
Retract +incompatible releases

bcmills:
(For #965.)

This retraction will take effect when this commit is included in the
latest v1 release (presumably v1.14.11).

PR #1015, opened 2022-02-17
Add explicit version instructions to README

eouw0o83hf:
**Background**

I am working on implementing ent in a codebase. [The testing page of ent](https://entgo.io/docs/testing/) specifically recommends using this repository.

`go get`ting this repository installs an invalid version which only fails in a way that's really difficult to analyze. Figuring out what the root cause was and figuring out how to install the correct version has eaten up an unfortunate chunk of my day.

**Change**

It looks like there's some other work going on in PRs to remediate the version issue. This doesn't pretend to fix that, but gives developers who wind up on this page a really explicit and easy way to solve that versioning problem before it wrecks their days.

Cheers!


rittneje:
@eouw0o83hf The v2 issue should have been resolved by #998. Now if I run `go get github.com/mattn/go-sqlite3` it chooses v1.14.15, as expected. Can this PR be declined?

PR #1017, opened 2022-02-22
Fix windows build with icu tag

codito:
In mingw64, [icu][] comes with `libicuin` instead of `libicui18n` library.

[icu]: https://packages.msys2.org/package/mingw-w64-x86_64-icu?repo=mingw64

### DIFF
--- a/sqlite3_opt_icu.go
+++ b/sqlite3_opt_icu.go
@@ -14,6 +14,6 @@ package sqlite3
 #cgo darwin CFLAGS: -I/usr/local/opt/icu4c/include
 #cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib -licui18n
 #cgo openbsd LDFLAGS: -lsqlite3 -licui18n
-#cgo windows LDFLAGS: -licuuc -licuin
+#cgo windows LDFLAGS: -licuin
 */
 import "C"
--- a/sqlite3_opt_icu.go
+++ b/sqlite3_opt_icu.go
@@ -8,10 +8,12 @@
 package sqlite3
 
 /*
-#cgo LDFLAGS: -licuuc -licui18n
+#cgo LDFLAGS: -licuuc
 #cgo CFLAGS: -DSQLITE_ENABLE_ICU
+#cgo linux LDFLAGS: -licui18n
 #cgo darwin CFLAGS: -I/usr/local/opt/icu4c/include
-#cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib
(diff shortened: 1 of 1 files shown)


sqlite3_opt_icu.go
```diff
@@ -8,10 +8,12 @@
...
-#cgo LDFLAGS: -licuuc -licui18n
+#cgo LDFLAGS: -licuuc
 #cgo CFLAGS: -DSQLITE_ENABLE_ICU
+#cgo linux LDFLAGS: -licui18n
 #cgo darwin CFLAGS: -I/usr/local/opt/icu4c/include
-#cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib
-#cgo openbsd LDFLAGS: -lsqlite3
+#cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib -licui18n
+#cgo openbsd LDFLAGS: -lsqlite3 -licui18n
+#cgo windows LDFLAGS: -licuuc -licuin
```
rittneje, line 17: I would assume that `-licuuc` here is redundant as it is unconditional above.

sqlite3_opt_icu.go
```diff
@@ -8,10 +8,12 @@
...
 /*
-#cgo LDFLAGS: -licuuc -licui18n
+#cgo LDFLAGS: -licuuc
 #cgo CFLAGS: -DSQLITE_ENABLE_ICU
+#cgo linux LDFLAGS: -licui18n
 #cgo darwin CFLAGS: -I/usr/local/opt/icu4c/include
-#cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib
-#cgo openbsd LDFLAGS: -lsqlite3
+#cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib -licui18n
+#cgo openbsd LDFLAGS: -lsqlite3 -licui18n
```
rittneje, line 16: @mattn Do you know why we are using `-lsqlite3` here? I would think that would cause issues since we'd be simultaneously dynamic linking against the system and statically compiling our own version.

mattn: Seems to be related on https://github.com/mattn/go-sqlite3/pull/595

rittneje: Hmm, I cannot test this since I don't have access to an openbsd system. And it's not clear if it was intentional. I will file a separate issue to investigate.

sqlite3_opt_icu.go
codito: @rittneje thanks you, fixed in latest commit.

PR #1019, opened 2022-02-23
Update amalgamation for 3.38.0

eatonphil:
https://sqlite.org/releaselog/3_38_0.html

### DIFF
--- a/sqlite3-binding.h
+++ b/sqlite3-binding.h
@@ -147,9 +147,9 @@ extern "C" {
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.37.0"
-#define SQLITE_VERSION_NUMBER 3037000
-#define SQLITE_SOURCE_ID      "2021-11-27 14:13:22 bd41822c7424d393a30e92ff6cb254d25c26769889c1499a18a0b9339f5d6c8a"
+#define SQLITE_VERSION        "3.38.0"
+#define SQLITE_VERSION_NUMBER 3038000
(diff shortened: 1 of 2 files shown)


rittneje:
> The JSON functions are now built-ins. It is no longer necessary to use the [-DSQLITE_ENABLE_JSON1](https://www.sqlite.org/compile.html#enable_json1) compile-time option to enable JSON support. JSON is on by default. Disable the JSON interface using the new [-DSQLITE_OMIT_JSON](https://www.sqlite.org/compile.html#omit_json) compile-time option.

@mattn As a followup I guess we can ditch sqlite3_opt_json1.go now.

sqlite3-binding.h
```diff
@@ -1,4 +1,3 @@
-#ifndef USE_LIBSQLITE3
```
rittneje, line 1: This has to be manually added again, or it won't work properly for people using the libsqlite3 build tag. Likewise for sqlite3-binding.c. This is due to how cgo works. https://pkg.go.dev/cmd/cgo

> When the Go tool sees that one or more Go files use the special import "C", it will look for other non-Go files in the directory and compile them as part of the Go package.

eatonphil: Oh thanks! I realized I should just use the upgrade tool. Which is very nice! Committed the result of running that tool.

mattn:
I would like to make it a rule that only I update the amagramation code, but since this PR has already been reviewed, so I would like to accept this PR this time.

eatonphil:
If it is or becomes no good, feel free to take it over or close it. Thanks!

mattn:
I'll make PR to remove sqlite3_opt_json1.go my-self in later. Thanks!

PR #1024, opened 2022-03-22
Update supported Golang version to Go 1.18

Aoang:


.github/workflows/go.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Get Build Tools
         run: |
-          GO111MODULE=on go get github.com/ory/go-acc
+          GO111MODULE=on go install github.com/ory/go-acc
```
mattn, line 32:
```suggestion
          GO111MODULE=on go install github.com/ory/go-acc@latest
```

mattn:
Thank you

PR #1035, opened 2022-04-23
Corrected RegisterPreUpdateHook Documentation

### LINKED ISSUE #1034: Question: RegisterPreUpdateHook Behavior
The documentation for [RegisterPreUpdateHook](https://pkg.go.dev/github.com/mattn/go-sqlite3#SQLiteConn.RegisterPreUpdateHook) states

> If there is an existing update hook for this connection, it will be removed. If callback is nil the existing hook (if any) will be removed without creating a new one.

This is the exact same verbiage as [RegisterUpdateHook](https://pkg.go.dev/github.com/mattn/go-sqlite3#SQLiteConn.RegisterUpdateHook).

Does this mean `RegisterPreUpdateHook` and `RegisterUpdateHook` cannot be used simultaneously? Or should the `RegisterPreUpdateHook` documentation actually say "If there is an existing **pre**update hook for this connection,..."?

Perhaps I'm misunderstanding these hooks, but I thought `RegisterPreUpdateHook` occurs before an `INSERT`/`UPDATE`/`DELETE` while `RegisterUpdateHook` happens after.

coreybutler:
This PR updates comments for the RegisterPreUpdateHook wrapper/connection methods to accurately reflect the SQLite behavior of this hook. Resolves #1034.

### DIFF
--- a/sqlite3_opt_preupdate_hook.go
+++ b/sqlite3_opt_preupdate_hook.go
@@ -33,7 +33,7 @@ import (
 // The callback is passed a SQLitePreUpdateData struct with the data for
 // the update, as well as methods for fetching copies of impacted data.
 //
-// If there is an existing update hook for this connection, it will be
+// If there is an existing preupdate hook for this connection, it will be
 // removed. If callback is nil the existing hook (if any) will be removed
 // without creating a new one.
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #1037, opened 2022-04-29
Add build tag sqlite_math_functions

nkt-dk:
Add build tag sqlite_math_functions to enable built-in scalar math functions using compiler flag SQLITE_ENABLE_MATH_FUNCTIONS. For reference: https://www.sqlite.org/lang_mathfunc.html.

I also updated the readme file, please let me know if you have any comments


lggruspe:
I think this extension requires linking to `math.h`:

```
#cgo LDFLAGS: -lm
```

nkt-dk:
Thank you for the feedback, I think you are correct. I updated the PR and added the linker flag

rittneje:
Please also update the GitHub merge gate to include this new tag in the "Tags: full" step.
https://github.com/mattn/go-sqlite3/blob/bedc2985efda510946f102c1e2a5595be070a2b7/.github/workflows/go.yaml#L46-L47

It would also be helpful to include a unit test (with the same build tag) to verify that the SQLite math functions are compiled in. I don't think it's necessary to cover all of them, just pick one to ensure the build options are being set properly. https://www.sqlite.org/lang_mathfunc.html

nkt-dk:
Hi @bearbin,
I did not have much time lately but @lggruspe continued this fix and create a new PR here: https://github.com/mattn/go-sqlite3/pull/1059 Thus I think my PR can be closed and we continue in 1059?

rittneje:
superseded by #1059

PR #1042, opened 2022-05-13
Upgrade Amalgamation to 3.38.5

bggwdc:
There are errors with certain queries with the new implementation of the
query planner in 3.38.0 that have been fixed with the 3.38.5 release.
Can we upgrade to this version?


mattn:
Thanks. I would like to make it a rule that only I update the amagramation code. I'll make new PR.

PR #1046, opened 2022-05-13
Support returning any from callbacks

### LINKED ISSUE #1045: Aggregator Done() can't return `any`/`interface{}`
Hey, I'm trying to write some aggregator functions that work on any data type. Here's a minimal example:

```
package main

import (
        "database/sql"
        "fmt"
        "log"

        sqlite "github.com/mattn/go-sqlite3"
)


type mode struct {
        counts   map[any]int
        top      any
        topCount int
}

func newMode() *mode {
        return &mode{
                counts: map[any]int{},
        }
}

func (m *mode) Step(x any) {
        m.counts[x]++
        c := m.counts[x]
        if c > m.topCount {
                m.top = x
                m.topCount = c
        }
}

func (m *mode) Done() any {
        return m.top
}

func main() {
        sql.Register("sqlite3_custom", &sqlite.SQLiteDriver{
                ConnectHook: func(conn *sqlite.SQLiteConn) error {
                        if err := conn.RegisterAggregator("mode", newMode, true); err != nil {
                                return err
                        }
                        return nil
                },
        })

        db, err := sql.Open("sqlite3_custom", ":memory:")
        if err != nil {
                log.Fatal("Failed to open database:", err)
        }
        defer db.Close()

        _, err = db.Exec("create table foo (department integer, profits integer)")
        if err != nil {
                log.Fatal("Failed to create table:", err)
        }
        _, err = db.Exec("insert into foo values (1, 10), (1, 20), (1, 45), (2, 42), (2, 115), (2, 20)")
        if err != nil {
                log.Fatal("Failed to insert records:", err)
        }

        rows, err := db.Query("select mode(profits) from foo")
        if err != nil {
                log.Fatal("MODE query error:", err)
        }
        defer rows.Close()
        for rows.Next() {
                //var dept int64
                var dev string
                if err := rows.Scan( &dev); err != nil {
                        log.Fatal(err)
                }
                fmt.Printf("mode=%s\n", dev)
        }
                log.Fatal(err)
        }
}
```

When I run this though I get a panic:

```
panic: reflect: Elem of invalid type interface {}

goroutine 1 [running]:
reflect.(*rtype).Elem(0x64b2a0?)
        /usr/local/go/src/reflect/type.go:965 +0x134
github.com/mattn/go-sqlite3.callbackRet({0x64bd60, 0x5fdfe0})
        /home/phil/tmp/sqlite3-test/vendor/github.com/mattn/go-sqlite3/callback.go:366 +0x17c
github.com/mattn/go-sqlite3.(*SQLiteConn).RegisterAggregator(0xc000076180, {0x61a96a, 0x4}, {0x5f6fa0, 0x625d60}, 0x1)
        /home/phil/tmp/sqlite3-test/vendor/github.com/mattn/go-sqlite3/sqlite3.go:776 +0xbaf
main.main.func1(0xc0000161a0?)
        /home/phil/tmp/sqlite3-test/main2.go:40 +0x39
github.com/mattn/go-sqlite3.(*SQLiteDriver).Open(0xc00006a040, {0x61b08b, 0x8})
        /home/phil/tmp/sqlite3-test/vendor/github.com/mattn/go-sqlite3/sqlite3.go:1765 +0x3c79
database/sql.dsnConnector.Connect(...)
        /usr/local/go/src/database/sql/sql.go:761
database/sql.(*DB).conn(0xc00010ea90, {0x64b7a0, 0xc00001e0e0}, 0x1)
        /usr/local/go/src/database/sql/sql.go:1395 +0x782
database/sql.(*DB).exec(0x4d770e?, {0x64b7a0, 0xc00001e0e0}, {0x6244c3, 0x36}, {0x0, 0x0, 0x0}, 0x40?)
        /usr/local/go/src/database/sql/sql.go:1657 +0x5d
database/sql.(*DB).ExecContext(0x61c4c1?, {0x64b7a0, 0xc00001e0e0}, {0x6244c3, 0x36}, {0x0, 0x0, 0x0})
        /usr/local/go/src/database/sql/sql.go:1635 +0xe5
database/sql.(*DB).Exec(...)
        /usr/local/go/src/database/sql/sql.go:1653
main.main()
        /home/phil/tmp/sqlite3-test/main2.go:53 +0x172
```

I changed `Done()` to return a string here and did:

```go
func (m *mode) Done() string {
        return fmt.Sprintf("%v", m.top)
}
```

And that works. But I don't want to force every result to a string here.

Are you open to patching the `callbackRet` function to support returning `any` from `Done()`? Or is that just not possible?

eatonphil:
Feel free to take this over or close this or make suggestions or whatever!

Closes #1045

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -354,7 +354,12 @@ func callbackRetNil(ctx *C.sqlite3_context, v reflect.Value) error {
 }
 
 func callbackRetGeneric(ctx *C.sqlite3_context, v reflect.Value) error {
-        cb, err := callbackRet(v.Elem().Type())
+	if v.IsNil() {
+		C.sqlite3_result_null(ctx)
+		return nil
+	}
+
+	cb, err := callbackRet(v.Elem().Type())
         if err != nil {
                 return err
         }
--- a/callback.go
+++ b/callback.go
@@ -353,7 +353,7 @@ func callbackRetNil(ctx *C.sqlite3_context, v reflect.Value) error {
 	return nil
 }
 
-func callbackRetAny(ctx *C.sqlite3_context, v reflect.Value) error {
+func callbackRetGeneric(ctx *C.sqlite3_context, v reflect.Value) error {
         cb, err := callbackRet(v.Elem().Type())
         if err != nil {
                 return err
@@ -371,7 +371,7 @@ func callbackRet(typ reflect.Type) (callbackRetConverter, error) {
 		}
 
 		if typ.NumMethod() == 0 {
-			return callbackRetAny, nil
+			return callbackRetGeneric, nil
 		}
 
 		fallthrough
--- a/callback.go
+++ b/callback.go
@@ -353,6 +353,15 @@ func callbackRetNil(ctx *C.sqlite3_context, v reflect.Value) error {
 	return nil
 }
 
+func callbackRetAny(ctx *C.sqlite3_context, v reflect.Value) error {
+        cb, err := callbackRet(v.Elem().Type())
+        if err != nil {
+                return err
+        }
+
+        return cb(ctx, v.Elem())
+}
+
 func callbackRet(typ reflect.Type) (callbackRetConverter, error) {
 	switch typ.Kind() {
 	case reflect.Interface:
@@ -360,6 +369,11 @@ func callbackRet(typ reflect.Type) (callbackRetConverter, error) {
 		if typ.Implements(errorInterface) {
 			return callbackRetNil, nil
 		}
+
+		if typ.NumMethod() == 0 {
+			return callbackRetAny, nil
+		}
+
 		fallthrough
 	case reflect.Slice:
 		if typ.Elem().Kind() != reflect.Uint8 {
--- a/callback_test.go
+++ b/callback_test.go
@@ -102,3 +102,15 @@ func TestCallbackConverters(t *testing.T) {
 		}
 	}
 }
+
+func TestCallbackReturnAny(t *testing.T) {
+	udf := func() interface{} {
+		return 1
+	}
+
+	typ := reflect.TypeOf(udf)
+	_, err := callbackRet(typ.Out(0))
+	if err != nil {
+		t.Errorf("Expected valid callback for any return type, got: %s", err)
+	}
+}
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -1449,6 +1449,63 @@ func TestAggregatorRegistration(t *testing.T) {
 	}
 }
 
+type mode struct {
+        counts   map[interface{}]int
+        top      interface{}
+        topCount int
+}
+
+func newMode() *mode {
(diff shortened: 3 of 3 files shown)


callback.go
```diff
@@ -353,13 +353,27 @@ func callbackRetNil(ctx *C.sqlite3_context, v reflect.Value) error {
...
 	switch typ.Kind() {
 	case reflect.Interface:
 		errorInterface := reflect.TypeOf((*error)(nil)).Elem()
 		if typ.Implements(errorInterface) {
 			return callbackRetNil, nil
 		}
+
+		if typ.NumMethod() == 0 {
+			return callbackRetAny, nil
+		}
```
eatonphil, line 375: I added a test for this part. But I'm not sure how to add a unit test for the callbackRetAny (I tested it manually end-to-end).

callback.go
```diff
@@ -353,13 +353,27 @@ func callbackRetNil(ctx *C.sqlite3_context, v reflect.Value) error {
 	return nil
 }
 
+func callbackRetAny(ctx *C.sqlite3_context, v reflect.Value) error {
```
rittneje, line 356: Probably should be `callbackRetGeneric` for consistency with `callbackArgGeneric`. Or rename the other one to `callbackArgAny`.

callback.go
rittneje: You can write a test that registers and uses such an aggregator, similar to [this test](https://github.com/mattn/go-sqlite3/blob/ae2a61f847e10e6dd771ecd4e1c55e0421cdc7f9/sqlite3_test.go#L1405).

callback.go
```diff
@@ -353,13 +353,27 @@ func callbackRetNil(ctx *C.sqlite3_context, v reflect.Value) error {
...
+        cb, err := callbackRet(v.Elem().Type())
```
rittneje, line 357: It is possible that the callback returned `nil`, in which case `v.Elem()` will return the invalid value. Presumably that needs to translate to SQLite `NULL`.

eatonphil: Do you mean I should have a check after the err check like

```
if cb == nil {
  return nil
}
```

?

callback.go
eatonphil: Perfect, thank you!

callback.go
eatonphil: Adjusted, thanks!

callback.go
rittneje: I think it would be this:
```go
if v.IsNil() {
    C.sqlite3_result_null(ctx)
    return nil
}

ve := v.Elem()
cb, err := callbackRet(ve.Type())
if err != nil {
    return err
}
return cb(ctx, ve)
```

eatonphil: Applied, thank you!

mattn:
@rittneje I'm okay to merge.

eatonphil:
Thanks folks!

PR #1048, opened 2022-05-14
&dest should be dest in pre update hook when getting old/new row

scaba:
Since the `dest` slice is already an address, taking the address of the address when updating the values causes the values to be lost in the caller


rittneje:
Thanks, can you also add a unit test that previously would have failed?

PR #1052, opened 2022-05-17
Add simple Dockerfile in "_example/simple" and its test

KEINOS:
This PR adds an example of building a `scratch`-based Docker image using a multi-stage build.

It also includes simple tests of containers built with GitHub Actions.

This is a PoC type, which means that if the correct compiler is used and the application is built statically and correctly, the built application binary will work correctly without additional installation.

This Dockerfile example may help isolate the problem for those struggling with 'CGO_ENABLED' and static build.

### DIFF
--- a/_example/simple/Dockerfile
+++ b/_example/simple/Dockerfile
@@ -1,0 +1,45 @@
+# =============================================================================
+#  Multi-stage Dockerfile Example
+# =============================================================================
+#  This is a simple Dockerfile that will build an image of scratch-base image.
+#  Usage:
+#    docker build -t simple:local . && docker run --rm simple:local
+# =============================================================================
+
+# -----------------------------------------------------------------------------
+#  Build Stage
+# -----------------------------------------------------------------------------
+FROM golang:alpine AS build
+
+# Important:
+#   Because this is a CGO enabled package, you are required to set it as 1.
+ENV CGO_ENABLED=1
+
+RUN apk add --no-cache \
+    # Important: required for go-sqlite3
+    gcc \
+    # Required for Alpine
+    musl-dev
+
+WORKDIR /workspace
+
+COPY . /workspace/
+
+RUN \
+    go mod init github.com/mattn/sample && \
+    go mod tidy && \
+    go install -ldflags="-s -w -extldflags \"-static\"" ./simple.go
+
+RUN \
+    # Smoke test
+    set -o pipefail; \
+    /go/bin/simple | grep 99\ こんにちわ世界099
+
+# -----------------------------------------------------------------------------
+#  Main Stage
+# -----------------------------------------------------------------------------
+FROM scratch
+
+COPY --from=build /go/bin/simple /usr/local/bin/simple
+
+ENTRYPOINT [ "/usr/local/bin/simple" ]
--- a/_example/simple/Dockerfile
+++ b/_example/simple/Dockerfile
@@ -28,7 +28,7 @@ COPY . /workspace/
 RUN \
     go mod init github.com/mattn/sample && \
     go mod tidy && \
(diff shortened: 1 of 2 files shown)


mattn:
Thank you. Could you please insert this?

```yaml

File: _example/simple/Dockerfile [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+# -----------------------------------------------------------------------------
+#  Build Stage
+# -----------------------------------------------------------------------------
+FROM golang:alpine AS build
+
+RUN apk add --no-cache \
+    # Build essentials for development on Alpine
+    alpine-sdk build-base \
+    # go-sqlite3 requirement
+    sqlite-dev
```
rittneje, line 18: I'm confused. This library statically links against its vendored copy of SQLite by default. I don't think this is required.

_example/simple/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+COPY . /workspace/
+
+RUN \
+    go mod init github.com/mattn/sample && \
+    go mod tidy && \
+    go install -ldflags="-s -w -extldflags \"-static\"" ./simple.go
+
+RUN \
+    # Smoke tests
+    which simple && \
```
rittneje, line 31: I don't think there's any need for the `which` command here. If the binary doesn't exist then invoking it will fail. On that note, I think it would be better to invoke it via the absolute path, to be sure you are calling the right thing.
Also, you should add `set -o pipefail` since by default `|` only fails if the rightmost command fails.

_example/simple/Dockerfile
KEINOS: Indeed. The minimum requirement is `gcc` (and `musl-dev` in Alpine).

My intention was to tell people to use build-essentials-for-dev for each distribution before issuing, as I have seen many issues with people forgetting to install the C compiler.

However, this seems to cause more confusion. I will slim it as much as possible.

_example/simple/Dockerfile
KEINOS: You got the point. I now realize that I am making a "if-then-else" case!

How about this?

```dockerfile
RUN /go/bin/simple | grep 99\ こんにちわ世界099
```

rittneje: You should still add `set -o pipefail` or else only the exit code from `grep` will be considered.

KEINOS: If `go/bin/simple` fails, then `grep` will fail because it cannot catch the "`99\ こんにちわ世界099`". Isn't it a bit redundant?

KEINOS: Never mind. Let's be cautious!

rittneje: It depends where it fails. If it fails after emitting that message (perhaps because it crashes), then without `pipefail` it would get masked.

KEINOS:
@mattn @rittneje

Done! (￣^￣)ゞ

Thanks for providing us such a cool package! Love this package and the community!

rittneje:
lgtm
@mattn?

_example/simple/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    musl-dev
+
+WORKDIR /workspace
+
+COPY . /workspace/
+
+RUN \
+    go mod init github.com/mattn/sample && \
+    go mod tidy && \
+    go install -ldflags="-s -w -extldflags \"-static\"" ./simple.go
```
mattn, line 31:
```suggestion
    go install -ldflags='-s -w -extldflags "-static'" ./simple.go
```

mattn:
LGTM. Will merge.

mattn:
Thank you

PR #1056, opened 2022-05-28
 Json example

mattn:
Now json is default. So I prefer to add _example/json to learn how to manipulate json in go-sqlite3.


rittneje:
@mattn The SQLite docs themselves include details about all the json functionality. https://www.sqlite.org/json1.html

As there isn't really anything interesting about this feature as far as this Go wrapper is concerned (like there is for, say, dealing with vtables), I would think that referring people to the official docs would be more useful in general.

mattn:
Sure

PR #1059, opened 2022-06-19
Add test for sqlite_math_functions tag

### LINKED ISSUE #990: Trying to use maths function (from SQLite v3.35) does not work
Since [this PR](https://github.com/mattn/go-sqlite3/pull/926) the underlying version of SQLite should be >=3.35 (and it is). The PR even mentions the new maths function added.
However, when I try to use math functions (like `SELECT power(2,2)`) I get this error `no such function: power`.

Is there something more I need to do than use the latest version of the go-sqlite3 library?

lggruspe:
Hello, this is a continuation of #1037, which fixes #990.

## Changes

- Adds test to check if sqlite math functions are compiled in
- Update github workflow to include sqlite_math_functions tag in full test

@rittneje @nkt-dk


README.md
```diff
@@ -172,6 +172,7 @@ go build --tags "icu json1 fts5 secure_delete"
 |  International Components for Unicode | sqlite_icu | This option causes the International Components for Unicode or "ICU" extension to SQLite to be added to the build |
 | Introspect PRAGMAS | sqlite_introspect | This option adds some extra PRAGMA statements. <ul><li>PRAGMA function_list</li><li>PRAGMA module_list</li><li>PRAGMA pragma_list</li></ul> |
 | JSON SQL Functions | sqlite_json | When this option is defined in the amalgamation, the JSON SQL functions are added to the build automatically |
+| Math Functions | sqlite_math_functions | This compile-time option enables built-in scalar math functions |
```
rittneje, line 175: Can you link to https://www.sqlite.org/lang_mathfunc.html? That way it's easy for people to see what all the math functions are.

sqlite3_opt_math_functions_test.go
```diff
@@ -0,0 +1,28 @@
...
+	defer db.Close()
+
+	queries := []string{
+		`SELECT acos(1)`,
+		`SELECT log(10, 100)`,
+		`SELECT power(2, 2)`,
+	}
+
+	for _, query := range queries {
+		if _, err := db.Exec(query); err != nil {
```
rittneje, line 24: You should use `db.QueryRow` for `SELECT`, not `db.Exec`.

sqlite3_opt_math_functions.go
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2019 Yasuhiro Matsumoto <mattn.jp@gmail.com>.
```
rittneje, line 1: @mattn What's the expectation for the copyright statement on new files?

README.md
lggruspe: Done

sqlite3_opt_math_functions_test.go
lggruspe: I also considered `QueryRow`, but it defers the error to `Scan`, and the result isn't really needed for the test. Maybe `Query` would be a better idea?

sqlite3_opt_math_functions.go
mattn: s/2019/2022/

sqlite3_opt_math_functions.go
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (C) 2018 G.J.R. Timmer <gjr.timmer@gmail.com>.
```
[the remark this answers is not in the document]
lggruspe: done

sqlite3_opt_math_functions_test.go
```diff
@@ -0,0 +1,28 @@
...
+	defer db.Close()
+
+	queries := []string{
+		`SELECT acos(1)`,
+		`SELECT log(10, 100)`,
+		`SELECT power(2, 2)`,
+	}
+
+	for _, query := range queries {
+		if _, err := db.Query(query); err != nil {
```
rittneje, line 24: Right now you are dropping the `Rows` object. That means you are only checking for an error from `sqlite3_prepare_v2` and not `sqlite3_step`. For completeness we should check both, which is why `QueryRow` would be preferable.

```go
var result float64
if err := db.QueryRow(query).Scan(&result); err != nil {
    t.Errorf("invoking math function query %q: %v", query, err)
}
// For the purposes of this test, we don't care about the output, just that the query succeeded.
```

lggruspe: I see... thanks for the review :)

mattn:
Seems okay to me.

mattn (approved):

rittneje:
Failure is due to unrelated issue introduced in #1085.

PR #1062, opened 2022-06-30
TestQueryer: actually check Rows returned

ohwgiles:
Fixes a test which did not correctly exercise the multi-statement Queryer functionality.

Note that the test now (correctly) fails because the implementation is broken.


mattn:
Thank you

Issue #1068, opened 2022-07-27
Native compilation with ICU for Apple Silicon fails


Spaider:
I'm relatively new to Golang and not sure if I get the whole picture. This is the reason I decided to create an issue but not MR yet.

I'm trying to compile *go-sqlite3* with ICU extension on Apple M1 chip. It fails because #cgo directive in [sqlite3_opt_icu.go](https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_icu.go) specifies lib and include directories under `/usr/local/opt` which is supposed to be Homebrew root. But this is not a case for M1/M2 chips. Homebrew uses `/opt/homebrew/opt` on those machines.

rittneje:
I believe you can change it like so:
```golang
#cgo darwin,amd64 CFLAGS: -I/usr/local/opt/icu4c/include
#cgo darwin,amd64 LDFLAGS: -L/usr/local/opt/icu4c/lib
#cgo darwin,arm64 CFLAGS: -I/opt/homebrew/opt/icu4c/include
#cgo darwin,arm64 LDFLAGS: -L/opt/homebrew/opt/icu4c/lib
```

mattn:
Fixed: darwin/arm64 ICU CFLAGS/LDFLAGS pointing at /opt/homebrew/opt/icu4c were added (PR #1069, sqlite3_opt_icu.go). Closing as resolved.

PR #1069, opened 2022-07-28
Add build tags to support both x86 and ARM compilation on macOS

Spaider:
Fix for #1068 

**Changes**

* Introduce `arm64` and `amd64` build tags
* Appropriate changes in documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This package follows the official [Golang Release Policy](https://golang.org/doc
     - [Alpine](#alpine)
     - [Fedora](#fedora)
     - [Ubuntu](#ubuntu)
-  - [Mac OSX](#mac-osx)
+  - [macOS X](#mac-osx)
   - [Windows](#windows)
   - [Errors](#errors)
 - [User Authentication](#user-authentication)
@@ -215,8 +215,8 @@ This library can be cross-compiled.
 
 In some cases you are required to the `CC` environment variable with the cross compiler.
 
-## Cross Compiling from MAC OSX
-The simplest way to cross compile from OSX is to use [xgo](https://github.com/karalabe/xgo).
+## Cross Compiling from macOS X
+The simplest way to cross compile from macOS X is to use [xgo](https://github.com/karalabe/xgo).
 
 Steps:
 - Install [xgo](https://github.com/karalabe/xgo) (`go get github.com/karalabe/xgo`).
@@ -267,9 +267,9 @@ sudo yum groupinstall "Development Tools" "Development Libraries"
 sudo apt-get install build-essential
 ```
 
-## Mac OSX
+## macOS X
 
-OSX should have all the tools present to compile this package. If not, install XCode to add all the developers tools.
+macOS X should have all the tools present to compile this package. If not, install XCode to add all the developers tools.
 
 Required dependency:
 
@@ -277,7 +277,7 @@ Required dependency:
 brew install sqlite3
 ```
 
-For OSX, there is an additional package to install which is required if you wish to build the `icu` extension.
+For macOS X, there is an additional package to install which is required if you wish to build the `icu` extension.
 
 This additional package can be installed with `homebrew`:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -288,13 +288,13 @@ brew upgrade icu4c
 To compile for macOS X on x86:
 
 ```bash
-go build --tags "darwin,amd64"
```
rittneje: These instructions were actually totally wrong. It should be using `GOOS` and `GOARCH` for this. The same applies for Windows and Linux. (Also the flag is `-tags` not `--tags`.)

README.md
```diff
@@ -40,7 +40,7 @@ This package follows the official [Golang Release Policy](https://golang.org/doc
     - [Alpine](#alpine)
     - [Fedora](#fedora)
     - [Ubuntu](#ubuntu)
-  - [Mac OSX](#mac-osx)
+  - [macOS X](#mac-osx)
```
rittneje, line 43: In modern times it is simply named `macOS`.

rittneje:
@mattn Do you know if GitHub has any M1/M2 build agents we could add to the workflow? https://github.com/mattn/go-sqlite3/blob/master/.github/workflows/go.yaml

mattn:
No I don't.

Spaider:
> @mattn Do you know if GitHub has any M1/M2 build agents

[Issue](https://github.com/actions/virtual-environments/issues/2187) is still open. Looks like it's in plans but not there yet.

README.md
Spaider: Sorry, I'm not following where's my fault here. I've added architecture flags exactly in a way it was for OS selection: by explicitly define target OS/arch. I'm a roocky in cgo and not sure how to re-write this to make it refer to `GOOS` and `GOARCH` env vars: `#cgo` flag does not have `if-then-else` notion.

rittneje: The original instructions (on master) were incorrect.

If you are doing a _native_ compile, it is just `go build`.
If you are doing a _cross_ compile, it is `GOOS=<target-os> GOARCH=<target-arch> CGO_ENABLED=1 CC=<c-cross-compiler> go build`
For example: `GOOS=darwin GOARCH=amd64 CGO_ENABLED=1 CC=o64-clang go build`

Separately from that, if you are trying to include optional features via build tags, in modern versions of Go you provide a comma-separated list via `-tags`.
> -tags tag,list
	a comma-separated list of build tags to consider satisfied during the
	build. For more information about build tags, see the description of
	build constraints in the documentation for the go/build package.
	(Earlier versions of Go used a space-separated list, and that form
	is deprecated but still recognized.)

For example: `go build -tags sqlite_stat4,secure_delete` 

Note that all of this is separate from the `#cgo` directive changes you made. (Those changes are correct.)

Spaider: Thanks for an explanation, now it makes perfect sense. Looks like the problem of both _native_ and _cross-compilcation_ can be resolved only by external tool. For example, shell script which will set `CGO_CGOCFLAGS` and `CGO_LDFLAGS` based on _target_ `GOOS` and `GOARCH`. But this is beyond the scope of my PR.

mattn (approved):

sqlite3_opt_icu.go
```diff
@@ -10,8 +10,10 @@ package sqlite3
 /*
 #cgo LDFLAGS: -licuuc -licui18n
 #cgo CFLAGS: -DSQLITE_ENABLE_ICU
-#cgo darwin CFLAGS: -I/usr/local/opt/icu4c/include
-#cgo darwin LDFLAGS: -L/usr/local/opt/icu4c/lib
+#cgo darwin,amd64 CFLAGS:  -I/usr/local/opt/icu4c/include
+#cgo darwin,amd64 LDFLAGS: -L/usr/local/opt/icu4c/lib
+#cgo darwin,arm64 CFLAGS:  -I/opt/homebrew/opt/icu4c/include
+#cgo darwin,arm64 LDFLAGS: -L/opt/homebrew/opt/icu4c/lib
```
mattn, line 16: I'm wondering why this is different for them.
```
-I/usr/local/opt/icu4c/include -I/opt/homebrew/opt/icu4c/include
-L/usr/local/opt/icu4c/lib -L/opt/homebrew/opt/icu4c/lib
```
This is not good?

rittneje: @mattn Not sure I understand your question, but Homebrew installs to a different location on arm64 (i.e, M1/M2) than on amd64 (i.e., Intel), as mentioned in the docs. https://docs.brew.sh/Installation

We could also choose to include _both_ directories for macOS regardless of GOARCH, and I suppose that would also work. (Although I think we should put /opt/homebrew first in that case.)

zzet:
Hey folks! I'm wondering if any progress there?

Spaider:
I've resolved the conflict in `README.md`.

sqlite3_opt_icu.go
ARolek: I just hit this issue on a M1 Mac and these locations are correct. This PR along with the build tags `go build -tags "libsqlite3 darwin arm64"` allowed me to build against the sqlite version installed via homebrew.

ARolek:
Just checking in. It looks like this PR was approved, but never merged. Is there anything blocking the merge at this time?

PR #1071, opened 2022-08-05
Rollback on constraint failure

joshbuddy:
This fixes an issue where deferred constraints can leave the transaction in an inconsistent state. In the event of a failed deferred constraint, we need to issue a rollback in order to make a transaction startable later.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -494,7 +494,7 @@ func (ai *aggInfo) Done(ctx *C.sqlite3_context) {
 // Commit transaction.
 func (tx *SQLiteTx) Commit() error {
 	_, err := tx.c.exec(context.Background(), "COMMIT", nil)
-	if err != nil && err.(Error).Code == C.SQLITE_BUSY {
+	if err != nil && (err.(Error).Code == C.SQLITE_BUSY || err.(Error).Code == C.SQLITE_CONSTRAINT) {
 		// sqlite3 will leave the transaction open in this scenario.
 		// However, database/sql considers the transaction complete once we
 		// return from Commit() - we must clean up to honour its semantics.
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -494,10 +494,12 @@ func (ai *aggInfo) Done(ctx *C.sqlite3_context) {
 // Commit transaction.
 func (tx *SQLiteTx) Commit() error {
 	_, err := tx.c.exec(context.Background(), "COMMIT", nil)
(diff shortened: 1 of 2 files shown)


mattn (approved):

mattn:
@rittneje Do you have an opinion?

rittneje:
Given that database/sql will not allow a transaction to be reused once you call `Commit()` (even if it fails), should we just issue a `ROLLBACK` on error, regardless of which specific error it was? The SQLite docs say this:
> If the transaction has already been rolled back automatically by the error response, then the ROLLBACK command will fail with an error, but no harm is caused by this.

joshbuddy:
Sounds good to me @rittneje. I've updated the PR and also updated the comment.

sqlite3.go
```diff
@@ -494,10 +494,12 @@ func (ai *aggInfo) Done(ctx *C.sqlite3_context) {
...
 func (tx *SQLiteTx) Commit() error {
 	_, err := tx.c.exec(context.Background(), "COMMIT", nil)
-	if err != nil && err.(Error).Code == C.SQLITE_BUSY {
-		// sqlite3 will leave the transaction open in this scenario.
+	if err != nil {
+		// sqlite3 may leave the transaction open in this scenario.
 		// However, database/sql considers the transaction complete once we
 		// return from Commit() - we must clean up to honour its semantics.
+		// We don't know if the ROLLBACK is strictly necessary, but according
+		// to sqlite's docs, there is no hard in calling ROLLBACK unnecessarily.
```
rittneje, line 502:
```suggestion
		// to sqlite's docs, there is no harm in calling ROLLBACK unnecessarily.
```

rittneje:
@joshbuddy LGTM, just one typo.

joshbuddy:
@rittneje fixed, thank you

joshbuddy:
Looks like some test flakiness?

rittneje:
@joshbuddy I think there may be something wrong with `TestQueryer`, and due to an oversight in how it got written, it is not reporting the real issue. Can you update that unit test like so?
```go
for rows.Next() {
    ...
}
    t.Fatal(err)
}
if n != 3 {
    ...
}
```

joshbuddy:
Hey @rittneje, I fixed the underlying test. If you want me to split it off into a separate PR, let me know.

joshbuddy:
I've moved the test fixes to https://github.com/mattn/go-sqlite3/pull/1079

rittneje (approved):

PR #1078, opened 2022-08-25
Add `PRAGMA trusted_schema` support

lberrymage:


infogulch:
I think this would be a valuable addition, and maybe should be the default. https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigtrustedschema

> When the SQLITE_DBCONFIG_TRUSTED_SCHEMA option is disabled, SQLite takes additional defensive steps to protect the application from harm including:
> 
> * Prohibit the use of SQL functions inside triggers, views, CHECK constraints, DEFAULT clauses, expression indexes, partial indexes, or generated columns unless those functions are tagged with [SQLITE_INNOCUOUS](https://www.sqlite.org/c3ref/c_deterministic.html#sqliteinnocuous).
> * Prohibit the use of virtual tables inside of triggers or views unless those virtual tables are tagged with [SQLITE_VTAB_INNOCUOUS](https://www.sqlite.org/c3ref/c_vtab_constraint_support.html#sqlitevtabinnocuous). 

In particular, sqlite devs recommend this option be on by default:

> This setting defaults to "on" for legacy compatibility, however **all applications are advised to turn it off if possible**. This setting can also be controlled using the [PRAGMA trusted_schema](https://www.sqlite.org/pragma.html#pragma_trusted_schema) statement.

rittneje:
@infogulch As it would be a breaking change, we cannot make this the default behavior.

infogulch:
That's fine, it should at least be an option.

lberrymage:
Because trusted_schema is per-connection, it can be enabled with a ConnectHook as follows:

```go
sql.Register(
    "sqlite3_hardened",
  		&sqlite3.SQLiteDriver{
				        _, err := conn.Exec("PRAGMA trusted_schema = OFF", nil)
				        return err
			     },
		  },
)
```

The database can then be opened as

```go
conn, err := sql.Open("sqlite3_hardened", "database.db")
```

infogulch:
I would prefer a connection string option. The ConnectHook option is annoying to use.

PR #1079, opened 2022-08-29
Fix test queryer test

joshbuddy:
This fixes the TestQueryer test to exec for the multistatement insertion, as well as fixing the value of n

### DIFF
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -1063,36 +1063,44 @@ func TestQueryer(t *testing.T) {
 	defer db.Close()
 
 	_, err = db.Exec(`
-	create table foo (id integer);
+		create table foo (id integer);
 	`)
 	if err != nil {
 		t.Error("Failed to call db.Query:", err)
 	}
 
-	rows, err := db.Query(`
-	insert into foo(id) values(?);
-	insert into foo(id) values(?);
-	insert into foo(id) values(?);
-	select id from foo order by id;
+	_, err = db.Exec(`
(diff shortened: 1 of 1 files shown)


sqlite3_test.go
```diff
@@ -1091,7 +1096,7 @@ func TestQueryer(t *testing.T) {
 		}
 		n = n + 1
 	}
-	if n != 3 {
+	if n != 4 {
```
mattn, line 1099: Why 4?

joshbuddy: It is a 1-indexed loop, so n is 4 when its looped three times. We could insert 0,1,2 and start at 0 instead, that way n would be 3 when its done. If you prefer that I'll make that change.

sqlite3_test.go
```diff
@@ -1091,7 +1096,7 @@ func TestQueryer(t *testing.T) {
 		}
 		n = n + 1
 	}
```
rittneje, line 1098: Please add the missing call to `rows.Err()` after the loop.

sqlite3_test.go
rittneje: I think it would be more intuitive to have `n` be the number of rows that were returned, in which case it should start at `0`. Whether you change it to insert 0, 1, 2, or change it to compare against `n+1` is fine to me.

joshbuddy:
@rittneje @mattn I've made those changes

rittneje (approved):

mattn:
Thank you

PR #1081, opened 2022-08-31
Fix #1080: vtable uses an explicit Omit flag now

### LINKED ISSUE #1080: VTab BestIndex sets .omit when Used=true
https://github.com/mattn/go-sqlite3/blob/master/sqlite3_opt_vtable.go#L467

I'm willing to submit a PR, but wanted to first see if this was viewed as a misinterpretation of the documentation. I got the impression that setting .omit is only for specifying that you will take care of the constraint and you're hinting to sqlite that it can skip the step if it wants.

From: [sqlite](https://www.sqlite.org/vtab.html#xfilter)
> it can try to suppress that double-check by setting aConstraintUsage[].omit

I guess I was hoping that one could set `.Used = true` in order to get hints and receive the right-side values in `Filter()`, but implement a best-effort filter and not a definite filter. 

Would we be open to changing the behavior here to say that `.Used = true` just means you want to use the constraint and you can still receive it in filter, and we can add a separate `.Omit = true` to be more explicit about the omit? Or maybe to preserve backward compat with existing behavior we can add something like `.UsedNoOmit = true`.

Thoughts?

cheezypoofs:
The virtual table omit flag is idenfified by sqlite as an optional
hint you can give the database about whether you are processed the
constraints, but the library doesn't require that you actually do
the filtering. Separting the .Used flag, which tells this go
wrapper to setup arguments to receive the constraints in Filter,
from the Omit flag allows us to inspect and perhaps do a best-effort
filtering and still allow sqlite proper to complete the evaluation
by leaving .omit false.

### DIFF
--- a/sqlite3_opt_vtable.go
+++ b/sqlite3_opt_vtable.go
@@ -3,6 +3,7 @@
 // Use of this source code is governed by an MIT-style
 // license that can be found in the LICENSE file.
 
+//go:build sqlite_vtable || vtable
 // +build sqlite_vtable vtable
 
 package sqlite3
@@ -367,6 +368,7 @@ func orderBys(info *C.sqlite3_index_info) []InfoOrderBy {
 // See: https://www.sqlite.org/c3ref/index_info.html
 type IndexResult struct {
 	Used           []bool // aConstraintUsage
+	Omit           []bool // sets the optional .omit flag
 	IdxNum         int
 	IdxStr         string
 	AlreadyOrdered bool // orderByConsumed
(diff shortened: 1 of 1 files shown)


mattn:
AFAICS, this seems to be breaking compatibility. @rittneje what do you think?

rittneje:
@mattn Yes, it is a breaking change. Previously `omit` was always set to `1` (i.e., `true`). Now it it will (presumably) default to `'0` (i.e., `false`) unless you set the new `Omit` field.

I think to properly preserve backwards compatibility, the condition needs to be `if res.Omit == nil || res.Omit[i]`. That way, all existing code (which will leave `Omit` as `nil`) will continue to set it as it does, and new code (which will assign `Omit` to a non-nil slice) can control it per index. Also, if `res.Omit != nil && len(res.Omit) != len(slice)` we should probably return an error (like we do for `res.Used`) as that seems like a developer mistake.

That said, the existing default logic for `omit` should be explicitly documented as it is the opposite of the default behavior in SQLite.

cheezypoofs:
Thanks for the consideration and feedback. I think I have reworked as you suggested.

rittneje:
@cheezypoofs Can you add a unit test that leverages the new functionality?

cheezypoofs:
Hmmm, I'll have to think of a way to properly exercise it. I'll give it a go.

Issue #239, opened 2015-09-16
Blob I/O


smhmw:
Would it be possible to implement the blob i/o API just as the backup API ?

[mxk/go-sqlite](https://github.com/mxk/go-sqlite/tree/master/sqlite3) has it but is a little out of date.
[gwenn/gosqlite](https://github.com/gwenn/gosqlite) has it too.

mattn:
database/sql doesn't provide Blob APIs. they provide Blob APIs for native driver (not database/sql).

smhmw:
Yes, but we can provide it aside from database/sql, like backup (backup.go) ? I may try to make it work locally, if it works i'll submit the patch - if that's interesting enough.

mattn:
No way to pass the Blob type into database.sql.

gwenn:
To make it work, you need access from `sql.DB` to the underlying/wrapped `sqlite3.SQLiteConn`...

smhmw:
I don't want to access it from `sql.DB`, rather add an API to the lib in a similar way to `sqlite3.SQLiteBackup`, such as `sqlite3.SQLiteBlob`; and corresponding method in `sqlite3.SQLiteConn` such as `Blob(db, table, column, rowid)` (similar to `SQLiteConn.Backup(...)`)

I'll give it a try ...

anacrolix:
:+1:

anacrolix:
Any progress on this @smhmw ?

GJRTimmer:
@smhmw Any update ?

ncruces:
I think I came up with a nice API for this in my SQLite driver, which could be implemented for this driver as well:
https://github.com/ncruces/go-sqlite3/blob/main/ext/blob/blob_test.go

This would require wrapping the [pointer passing](https://www.sqlite.org/bindptr.html) APIs, on top of blob IO.

PR #1083, opened 2022-09-01
Add Blob I/O

joriszwart:
Incremental Blob I/O
=====================

This PR adds support for [Incremental Blob I/O](https://www.sqlite.org/c3ref/blob_open.html).

Notes
-----

- Modeled after [backup](/mattn/go-sqlite3/blob/master/backup.go) using a driver connection.
- The interfaces *io.Reader* and *io.Closer* have been implemented.
- Related to issue #239

If this PR makes sense, I'll enhance it with additional *io* interfaces.

### DIFF
--- a/sqlite3_opt_column_metadata.go
+++ b/sqlite3_opt_column_metadata.go
@@ -6,7 +6,7 @@ package sqlite3
 /*
 #ifndef USE_LIBSQLITE3
 #cgo CFLAGS: -DSQLITE_ENABLE_COLUMN_METADATA
-#include <sqlite3-binding.h>
+#include "sqlite3-binding.h"
 #else
 #include <sqlite3.h>
 #endif
--- a/sqlite3_opt_serialize.go
+++ b/sqlite3_opt_serialize.go
@@ -5,7 +5,7 @@ package sqlite3
 
 /*
 #ifndef USE_LIBSQLITE3
-#include <sqlite3-binding.h>
+#include "sqlite3-binding.h"
 #else
 #include <sqlite3.h>
 #endif


blob_io.go
```diff
@@ -0,0 +1,94 @@
...
+	tableptr := C.CString(table)
+	defer C.free(unsafe.Pointer(tableptr))
+
+	columnptr := C.CString(column)
+	defer C.free(unsafe.Pointer(columnptr))
+
+	var blob *C.sqlite3_blob
+	ret := C.sqlite3_blob_open(conn.db, databaseptr, tableptr, columnptr, C.longlong(rowid), C.int(flags), &blob)
+
+	if ret == C.SQLITE_OK {
```
rittneje, line 48: Conventionally you should always handle the error case first, so this should be `if ret != C.SQLITE_OK`.

blob_io_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"io"
+	"testing"
+)
+
+func TestBlobIO(t *testing.T) {
+	driverName := "sqlite3_TestBlobIO2"
+	var driverConn *SQLiteConn
+	sql.Register(driverName, &SQLiteDriver{
+		ConnectHook: func(conn *SQLiteConn) error {
+			driverConn = conn
```
rittneje, line 23: Please use [`Conn.Raw`](https://pkg.go.dev/database/sql#Conn.Raw) instead, as that is also what we should be suggesting actual clients use.

blob_io.go
```diff
@@ -0,0 +1,94 @@
...
+
+// SQLiteBlob implements the SQLite Blob I/O interface.
+type SQLiteBlob struct {
+	conn *SQLiteConn
+	blob *C.sqlite3_blob
+	size int
+	offs int
+}
+
+// Blob opens a blob.
```
rittneje, line 32: Please fill this out with more details. Linking to the relevant SQLite3 docs will also be helpful.

blob_io.go
```diff
@@ -0,0 +1,94 @@
...
+
+	columnptr := C.CString(column)
+	defer C.free(unsafe.Pointer(columnptr))
+
+	var blob *C.sqlite3_blob
+	ret := C.sqlite3_blob_open(conn.db, databaseptr, tableptr, columnptr, C.longlong(rowid), C.int(flags), &blob)
+
+	if ret == C.SQLITE_OK {
+		size := int(C.sqlite3_blob_bytes(blob))
+		bb := &SQLiteBlob{conn, blob, size, 0}
```
rittneje, line 50:
```suggestion
		bb := &SQLiteBlob{conn: conn, blob: blob, size: size, offset: 0}
```

blob_io.go
```diff
@@ -0,0 +1,94 @@
...
+	"runtime"
+	"unsafe"
+)
+
+// SQLiteBlob implements the SQLite Blob I/O interface.
+type SQLiteBlob struct {
+	conn *SQLiteConn
+	blob *C.sqlite3_blob
+	size int
+	offs int
```
rittneje, line 29:
```suggestion
	offset int
```

blob_io.go
```diff
@@ -0,0 +1,147 @@
...
+		abs = int64(s.size) + offset
+	default:
+		return 0, errors.New("sqlite3.SQLiteBlob.Seek: invalid whence")
+	}
+
+	if abs < 0 {
+		return 0, errors.New("sqlite.SQLiteBlob.Seek: negative position")
+	}
+
+	s.offset = int(abs)
```
rittneje, line 130: This conversion can overflow on a 32-bit machine.

blob_io.go
```diff
@@ -0,0 +1,147 @@
...
+	return n, nil
+}
+
+// Write implements the io.Writer interface.
+func (s *SQLiteBlob) Write(b []byte) (n int, err error) {
+	if s.offset >= s.size {
+		return 0, io.EOF
+	}
+
+	n = s.size - s.offset
```
rittneje, line 92: Currently you are not meeting the `io.Writer spec. In particular:
> Write must return a non-nil error if it returns n < len(p).

blob_io.go
```diff
@@ -0,0 +1,147 @@
...
+
+	s.offset += n
+
+	return n, nil
+}
+
+// Write implements the io.Writer interface.
+func (s *SQLiteBlob) Write(b []byte) (n int, err error) {
+	if s.offset >= s.size {
+		return 0, io.EOF
```
rittneje, line 89: I'm not sure `io.EOF` is appropriate here. That is intended to be returned from `Read` not `Write`.

blob_io.go
```diff
@@ -0,0 +1,147 @@
...
+	var abs int64
+	switch whence {
+	case io.SeekStart:
+		abs = offset
+	case io.SeekCurrent:
+		abs = int64(s.offset) + offset
+	case io.SeekEnd:
+		abs = int64(s.size) + offset
+	default:
+		return 0, errors.New("sqlite3.SQLiteBlob.Seek: invalid whence")
```
rittneje, line 123: Can we use `fmt.Errorf` instead so we can report what `whence` we got? That should help them debug.

blob_io.go
rittneje: One thing that is currently missing is you haven't provided a way for the client to ask how long the blob is. Consequently there is no way for them to attempt to avoid encountering this error unless they already know through some other means. Can you add a `Size() int` method?

With that in place, I'm not sure whether client's will need to check this error with `errors.Is` or `errors.As`, so you can probably just do something like `fmt.Errorf("sqlite3: insufficient space in %d-byte blob", s.size)`.

blob_io.go
rittneje: https://www.sqlite.org/limits.html

The default limit on a blob is one billion bytes. It could be increased (particularly if you are using the `libsqlite3` build tag) but anything beyond 2^31 is "unsupported". So enforcing this check seems reasonable.

That said, should we instead enforce that it does not exceed `s.size` for consistent behavior? I'm not sure what the use case would be for seeking beyond the end of the blob, given that you can neither read nor write in that state.

blob_io.go
rittneje: Yeah I think an error consistent with the one you return for a short write would be good. For example: `fmt.Errorf("sqlite3: no space left in %d-byte blob", s.size)`.

blob_io_test.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+// Verify interface implementations
+var _ io.Reader = &SQLiteBlob{}
+var _ io.Writer = &SQLiteBlob{}
+var _ io.Seeker = &SQLiteBlob{}
+var _ io.Closer = &SQLiteBlob{}
+
+func blobTestData(dbname string, rowid int64, blob []byte) (*sql.DB, *SQLiteConn, error) {
+	db, err := sql.Open("sqlite3", "file:"+dbname+"?mode=memory&cache=shared")
```
rittneje, line 26: As shared cache mode is discouraged by the SQLite authors, I would prefer not to use it in tests so that clients don't get the wrong idea. Instead please use `file:/<dbname>?vfs=memdb`. Note the leading slash in the database name - this is critical for a memdb to be shared across connections in the pool.

blob_io_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	conn, err := db.Conn(context.Background())
+	if err != nil {
+		return nil, nil, err
+	}
+	defer conn.Close()
+
+	var driverConn *SQLiteConn
+	err = conn.Raw(func(conn interface{}) error {
+		driverConn = conn.(*SQLiteConn)
```
rittneje, line 39: You should not do this. It is a violation of `conn.Raw` to reference the connection outside the callback.
For your tests to work properly, please change `blobTestData` to take a callback `func(*testing.T, *SQLiteConn)`.

blob_io_test.go
rittneje: If you use `:memory:` then you can run into issues due to the connection pooling, as each connection will be using a separate in-memory database. For this reason, `vfs=memdb` is generally preferred.

blob_io_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+// Verify interface implementations
+var _ io.Reader = &SQLiteBlob{}
+var _ io.Writer = &SQLiteBlob{}
+var _ io.Seeker = &SQLiteBlob{}
+var _ io.Closer = &SQLiteBlob{}
+
+type driverConnCallback func(*testing.T, *SQLiteConn)
+
+func blobTestData(t *testing.T, dbname string, rowid int64, blob []byte, c driverConnCallback) error {
```
rittneje, line 27: Rather than returning an error, you can just call `t.Fatal` in here.

blob_io_test.go
```diff
@@ -0,0 +1,260 @@
...
+	err = conn.Raw(func(conn interface{}) error {
+		driverConn = conn.(*SQLiteConn)
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	defer driverConn.Close()
+
+	c(t, driverConn)
```
rittneje, line 66: This needs to happen inside the `conn.Raw` callback.

mattn:
@rittneje are you okay to merge?

blob_io.go
```diff
@@ -0,0 +1,161 @@
...
+		return 0, fmt.Errorf("sqlite3.SQLiteBlob.Write: insufficient space in %d-byte blob", s.size)
+	}
+
+	n = s.size - s.offset
+	if len(b) < n {
+		n = len(b)
+	}
+
+	if n != len(b) {
+		return n, fmt.Errorf("sqlite3.SQLiteBlob.Write: insufficient space in %d-byte blob", s.size)
```
rittneje, line 99: This violates the `io.Writer` interface. In particular:
> It returns the number of bytes written from p

Since you did not actually write anything to the blob, this needs to return 0. Alternatively, do this check after the call to `sqlite3_blob_write`.

blob_io.go
```diff
@@ -0,0 +1,161 @@
...
+		abs = int64(s.size) + offset
+	default:
+		return 0, fmt.Errorf("sqlite3.SQLiteBlob.Seek: invalid whence %d", whence)
+	}
+
+	if abs < 0 {
+		return 0, errors.New("sqlite.SQLiteBlob.Seek: negative position")
+	}
+
+	if abs > math.MaxInt32 {
```
rittneje, line 135: Can we instead check if `abs > int64(s.size)`?

blob_io.go
```diff
@@ -0,0 +1,161 @@
...
+
+	s.offset += n
+
+	return n, nil
+}
+
+// Seek implements the io.Seeker interface.
+func (s *SQLiteBlob) Seek(offset int64, whence int) (int64, error) {
+	if offset > math.MaxInt32 {
+		return 0, errors.New("sqlite3.SQLiteBlob.Seek: invalid position")
```
rittneje, line 116:
```suggestion
		return 0, fmt.Errorf("sqlite3.SQLiteBlob.Seek: invalid offset %d", offset)
```

blob_io_test.go
rittneje: This was not done.

graf0:
Hello!

> The Linux builds with _libsqlite3_ fail with:
> 
> ```
> === RUN   TestBlobRead
>     blob_io_test.go:46: no such vfs: memdb
> --- FAIL: TestBlobRead (0.00s)
> ```
> 
> I don't know how to fix this.

Problem is like this:

* github actions uses ubuntu-latest, which is focal release, version 20.04 (stable release) at time of writing this answer
* default version of sqlite3 in this release of ubuntu is 3.31.0
* it's compiled without SQLITE_ENABLE_DESERIALIZE which is required to enable memdb vfs in sqlite verison < 3.36.0

So - vfs=memdb will not work in ubuntu focal. You would need to recompile libsqlite3 deb package and add SQLITE_ENABLE_DESERIALIZE option to make. Without it memdb is just not there. And package libsqlite3-dev do not consist c code of sqlite3 - it's .a and .so files, already precompiled.

This behaviour was changed in sqlite 3.36.0 - SQLITE_ENABLE_DESERIALIZE is defined by default, you need to explicity disable it by defining SQLITE_OMIT_DESERIALIZE option during sqlite compilation.

So, there are imho there are two solution of this problem:

* use connection string: "file:foobar?mode=memory&cache=shared" - maybe only if version of sqlite3 is < 3.36.0
* or use connection string ":memory:" and set db.SetMaxOpenConns(1) (mayb followed by db.Ping() to create conn?) to limit all access to the same memory database on all versions of sqlite, including sqlite3

This problem will disappear after ubuntu will release next LTS version, and github action will change ubuntu-latest to this LTS version - there should be at least sqlite3 3.37.0 there.

graf0:
sugested fix (will work with any version of sqlite):

```diff
diff --git a/blob_io_test.go b/blob_io_test.go
index 3e6fb91..b98df8d 100644
--- a/blob_io_test.go
+++ b/blob_io_test.go
@@ -25,12 +25,14 @@ var _ io.Closer = &SQLiteBlob{}
 type driverConnCallback func(*testing.T, *SQLiteConn)
 
 func blobTestData(t *testing.T, dbname string, rowid int64, blob []byte, c driverConnCallback) {
-       db, err := sql.Open("sqlite3", "file:/"+dbname+"?vfs=memdb")
+       db, err := sql.Open("sqlite3", ":memory:")
        if err != nil {
                t.Fatal(err)
        }
        defer db.Close()
 
+       db.SetMaxOpenConns(1)
+
        // Test data
        query := `
                CREATE TABLE data (
```

rittneje:
@joriszwart In this particular case `:memory:` should work. However, the need to throttle the connection pool to one can cause some issues in general. If you make that change, you should also add comments explaining it is specifically to support older SQLite versions and developers should always use `file:/<name>?vfs=memdb` instead whenever possible. (Developers may reference these test cases to see how to use the blob i/o feature.)

Really I think we need a better approach for dealing with testing `libsqlite3` in general. But that is an issue beyond the scope of your changes.

jlelse:
Any updates on this? It looks very interesting and enables "streaming" to and from the database! 👍

blob_io.go
```diff
@@ -0,0 +1,161 @@
...
+	if s.offset >= s.size {
+		return 0, io.EOF
+	}
+
+	n = s.size - s.offset
+	if len(b) < n {
+		n = len(b)
+	}
+
+	p := &b[0]
```
rittneje, line 76: This will panic if `len(b) == 0`, which is technically legal. We should probably handle that explicitly to return `0, nil`, after the `io.EOF` handling.

blob_io.go
```diff
@@ -0,0 +1,161 @@
...
+	n = s.size - s.offset
+	if len(b) < n {
+		n = len(b)
+	}
+
+	if n != len(b) {
+		return 0, fmt.Errorf("sqlite3.SQLiteBlob.Write: insufficient space in %d-byte blob", s.size)
+	}
+
+	p := &b[0]
```
rittneje, line 102: This also panics if `len(b) == 0`. That should return `0, nil`.

blob_io.go
```diff
@@ -0,0 +1,161 @@
...
+	if s.offset >= s.size {
+		return 0, fmt.Errorf("sqlite3.SQLiteBlob.Write: insufficient space in %d-byte blob", s.size)
+	}
+
+	n = s.size - s.offset
+	if len(b) < n {
+		n = len(b)
+	}
+
+	if n != len(b) {
```
rittneje, line 98: Trying to remember - is there a reason not to do this check after the call to `sqlite3_blob_write` and only write what we can instead of nothing?

rittneje:
@joriszwart I am away from my dev machine or I'd just add those two `len` checks myself. They should be pretty simple to add.

blob_io.go
rittneje: @joriszwart following up on this

blob_io_test.go
```diff
@@ -0,0 +1,252 @@
...
+var _ io.Reader = &SQLiteBlob{}
+var _ io.Writer = &SQLiteBlob{}
+var _ io.Seeker = &SQLiteBlob{}
+var _ io.Closer = &SQLiteBlob{}
+
+type driverConnCallback func(*testing.T, *SQLiteConn)
+
+func blobTestData(t *testing.T, dbname string, rowid int64, blob []byte, c driverConnCallback) {
+
+	// TODO use :memory: for compatibility with SQLite versions < 3.37.0.
```
rittneje, line 29: @joriszwart please resolve this TODO

blob_io_test.go
```diff
@@ -0,0 +1,252 @@
...
+	var driverConn *SQLiteConn
+	err = conn.Raw(func(conn interface{}) error {
+		driverConn = conn.(*SQLiteConn)
+		c(t, driverConn)
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer driverConn.Close()
```
rittneje, line 72: Remove. This is superfluous with the call to `conn.Close()` above.

gabriel-samfira:
A shame this was abandoned. Thanks for trying @joriszwart !

gabriel-samfira:
> @mattn @rittneje any interest in this?

for what it's worth, I would love to see this happen.

blob_io_test.go
gabriel-samfira: I think just line 72. The `driverConn` is just `conn` cast as the underlying type that implements the interface (`conn`). Calling Close() on `conn` will close `driverConn` as well.

blob_io.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	s.offset += n
+
+	return n, nil
+}
+
+// Write implements the io.Writer interface.
+func (s *SQLiteBlob) Write(b []byte) (n int, err error) {
+	if len(b) == 0 {
```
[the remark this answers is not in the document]
gabriel-samfira: Not really, as you'd only be copying the buffer (which is small - usually around 1024 bytes). But generally speaking you'd only have to copy if you need to guard against caller reuse of the buffer. This is usually done in logging writers that may want to return control back to the caller while the write operation against the logging backend is still ongoing.

I'm not sure that would apply here though (given that we want to wait for the write to happen and only then return), so I just added a comment as a "recommendation". Feel free to ignore me.

I haven't mentioned this, but thanks for reopening this PR!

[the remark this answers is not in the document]
rittneje: There is no need to make a copy of the slice, as `sqlite3_blob_write` is going to copy the data. And the caller is not allowed to modify the slice _during_ the call the `Write`. (Even if they did, `copy` itself would be subject to a race condition anyway.)

blob_io.go
gabriel-samfira: You're right on both counts.

gabriel-samfira:
for what it's worth, I rebased this PR on top of main and I'm using it in one of my projects. It works great. This actively makes my life easier, as I don't need to add an additional requirement for users to configure a new writable path for small (very limited in number and size) files that the app requires to function properly. So thanks @joriszwart for this PR.

joriszwart:
So, how to proceed?

PR #1084, opened 2022-09-01
remove unuseful ldflags for windows platform.

KuoKongQingYun:
Golang's linker add mingwex and mingw32 automatically,so we don't need add them manually.


KuoKongQingYun:
I explain why I need remove thoes ldflags.
My project is dependent on  a library built by MSVC,so I use Clang for msvc to build  with cgo,but linker of MSVC don't has mingwex and mingw32.

rittneje:
LGTM but I am not a Windows developer. @mattn any concerns?

mattn:
Thank you.

PR #1089, opened 2022-09-06
Add Serialize and Deserialize support

otoolep:


otoolep:
It's not clear to me why the build is failing because:

- `sqlite3_serialize()` is available in the C source.
- SQLITE_OMIT_DESERIALIZE is not defined anywhere that I can see so the C function should be available.

What am I missing?

rittneje:
@otoolep Prior to v3.36.0, `sqlite3_serialize` and `sqlite3_deserialize` were opt in via `-DSQLITE_ENABLE_DESERIALIZE`. This is why the `libsqlite3` tests are failing. Also now they are opt out via `-DSQLITE_OMIT_DESERIALIZE`, which we should support.

We will either need a Go build tag to enable these new methods, or a tag to disable them. Personally, my preference would be to make it opt out so things "just work" by default, but technically that could be a breaking change for anyone currently using the `libsqlite3` tag. @mattn what do you think?

mattn:
>Personally, my preference would be to make it opt out so things "just work" by default

Do you mean that the functions return like ErrNotImplemented?

rittneje:
@mattn Historically this library has been inconsistent between having alternate version of the Go methods that return errors, or simply not defining the Go methods at all. I don't care which one we do here. The important part is that if you opt out (or don't opt in, depending on our approach), we never attempt to reference the C functions.

otoolep:
I can add a build tag for either option (opt-in or opt-out), I forgot about the `libsqlite3` tests. The lack of this patch is the only thing forcing rqlite to use a fork of this repo.

mattn:
```
type Serializable interface {
  Serialize(schema string) []byte
}
```
```
if ser, ok := conn.(Serializable); ok {
  b = ser.Serialize(name)
  // blah
} else {
  // not implemented
}
```
We can do check whether the SQLiteConn has methods like above, I think. @rittneje what do you think?

rittneje:
@mattn The underlying problem is we cannot reference the `sqlite3_serialize` or `sqlite3_deserialize` C functions if SQLite itself was compiled without them. For the statically compiled version this is simple because we control the build options, but for someone linking against their system version we have no way of knowing unless they tell us. There are the choices I can see:

1. We make our references to those two C functions opt out. This means anyone using `libsqlite3` today could be broken (if their system version was compiled without support) and will have to add an additional Go build tag so we don't attempt to reference those functions.
2. We make our references to those C functions opt in. This preserves backwards compatibility for anyone using `libsqlite3` today, but in exchange anyone that wants to use these new Go methods is required to pass an additional tag to `go build`, which is annoying.
3. We always exclude these new methods if `libsqlite3` is defined.
4. We make these new methods opt in if you are using `libsqlite3`. Otherwise they are always defined.
5. We make these new methods opt in if you are using `libsqlite3`, and opt out if you aren't.

Us dynamically checking what methods `SQLiteConn` has does not help, since we are the ones that have to define those very methods in the first place.

mattn:
Thanks your explaining. As you mentioned above, let's add `sqlite_omit_deserialize`.

rittneje:
To be clear, you are saying to do option 1 from my list?

mattn:
I have thought about this for a little while and I expect the following.

1. opt-in if using libsqlite3, default is undefined
2. always enabled if libsqlite3 is not used

mattn:
I assume that people who use libsqlite3 are expecting minimalism and they does not want these functions. I guess.

rittneje:
> I assume that people who use libsqlite3 are expecting minimalism and they does not want these functions. I guess.

Depends on their use case. They could also just want consistency across applications written in multiple languages.

otoolep:
@rittneje @mattn -- does this look like what you had in mind?

I also don't understand the "dockerfile" CI failure. Can you help?

rittneje:
> I also don't understand the "dockerfile" CI failure. Can you help?

Seems that #1085 changed the output of the sample program, but did not change the expectation in the GitHub action, so it fails. https://github.com/mattn/go-sqlite3/blob/7476442ed657d21b68b04e6ba2ae173f98265cf8/.github/workflows/docker.yaml#L22

That said, I don't know why the output is checked in both the Dockerfile itself as well as the GitHub action. @mattn Can one of them be removed?

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,68 @@
+// +build !sqlite_omit_deserialize
```
rittneje, line 1: This is not quite right. This file should be used if `!libsqlite3` or if `sqlite_serialize` (or both). That way it is opt-in for anyone linking with their system version, and always included otherwise. https://github.com/mattn/go-sqlite3/pull/1089#issuecomment-1238808014

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,68 @@
...
+*/
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+// Serialize returns a byte slice that is a serialization of the database.
+// If the database fails to serialize, a nil slice will be returned.
```
rittneje, line 22: Can we make it return an error in this case to play more nicely with the Go ecosystem? Otherwise clients that don't read this carefully might think it always succeeds.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,68 @@
...
+	zSchema = C.CString(schema)
+	defer C.free(unsafe.Pointer(zSchema))
+
+	var sz C.sqlite3_int64
+	ptr := C.sqlite3_serialize(c.db, zSchema, &sz, 0)
+	if ptr == nil {
+		return nil
+	}
+	defer C.sqlite3_free(unsafe.Pointer(ptr))
+	return C.GoBytes(unsafe.Pointer(ptr), C.int(sz))
```
rittneje, line 39: Casting `sqlite3_int64` to `int` could overflow on some systems.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Deserialize causes the connection to disconnect from the current database
+// and then re-open as an in-memory database based on the contents of the
+// byte slice. If deserelization fails, error will contain the return code
+// of the underlying SQLite API call.
+//
+// When this function returns, the connection is referencing database
```
rittneje, line 47: This is a violation of the rules of cgo. https://pkg.go.dev/cmd/cgo#hdr-C_references_to_Go

> C code may not keep a copy of a Go pointer after the call returns.

mattn:
> 

please substitute `こんにちわ世界099` to `こんにちは世界099`

sqlite3_opt_serialize.go
otoolep: >This is not quite right. This file should be used if !libsqlite3 or if sqlite_serialize (or both). 

I presume you mean `sqlite_omit_deserialize` since I was asked to add it in https://github.com/mattn/go-sqlite3/pull/1089#issuecomment-1238743103

sqlite3_opt_serialize.go
otoolep: Thank you for pointing that out.

otoolep: Actually, why is it an issue? I presume you're talking about the `b` that is passed in -- but its lifetime is managed by the *caller* for this function. The caller will control when it goes out of scope. Or am I missing something? To be clear, I want to change this, the current implementation is clumsy, but want to be sure I understand the issue.

Do I first have to copy the Go slice to C memory space (or something like that?).

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,68 @@
+// +build libsqlite3 !sqlite_omit_deserialize
```
otoolep, line 1: @rittneje -- is this what you want?

sqlite3_opt_serialize_test.go
```diff
@@ -0,0 +1,119 @@
...
+	defer conn.Close()
+	if err := sqlite3conn.Deserialize(b, ""); err != nil {
+		t.Fatal("failed to deserialize database", err)
+	}
+
+	// Check database contents are as expected.
+	rows, err := sqlite3conn.Query(`SELECT * FROM foo`, nil)
+	if err != nil {
+		t.Fatal("failed to query database:", err)
+	}
```
mattn, line 108:
```
defer rows.Close()
```

sqlite3_opt_serialize.go
otoolep: @rittneje -- since the signature of `GoBytes` requires `C.int` should I copy the data byte-by-byte in a loop?

sqlite3_opt_serialize_test.go
otoolep: Fixed.

sqlite3_opt_serialize.go
rittneje: See the subsequent comment from @mattn. https://github.com/mattn/go-sqlite3/pull/1089#issuecomment-1238808014

sqlite3_opt_serialize.go
rittneje: Unless the caller explicitly does something to force it to remain in scope (e.g., `runtime.KeepAlive`) it can actually be garbage collected immediately. And even if they do (or you added `runtime.KeepAlive` here), it is still a violation as the Go runtime reserves the right to move stuff around in memory (right now it doesn't but may in the future).

I think you will need to first copy the Go `[]byte` into memory obtained via `sqlite3_malloc`. Then you can pass `SQLITE_DESERIALIZE_FREEONCLOSE` to to `sqlite3_deserialize`. I cannot think of any other way to make it truly safe.

sqlite3_opt_serialize.go
otoolep: Sorry, I'm not following. Can you be more explicit? What build tags do you want added, exactly?

sqlite3_opt_serialize.go
rittneje: Since Go represents the length (and capacity) of a slice as an `int`, the potential overflow will be an issue on 32-bit systems no matter what. Thus in that case you have no choice but to return an error. However, on a 64-bit system, you should be safe.

Regardless, there is no guarantee about `C.int` itself on either system. It is _likely_ to be 32-bits in both cases, but the C standard has no strict requirement.

I would not copy byte-by-byte as that will be very inefficient, particularly for large databases. In the interest of keeping things simple (at least for your initial implementation), I think we can just do some checking around `C.sizeof_int` like so:
```go
if C.sizeof_int < 64 {
    maxSize := C.sqlite3_int64(1) << C.sizeof_int - 1
    if sz > maxSize {
        return nil, fmt.Errorf("sqlite3: serialized database is too large (%d bytes)", maxSize)
    }
}
return C.GoBytes(unsafe.Pointer(ptr), C.int(sz)), nil
```

If we wanted to support larger databases on 64-bit systems, I think it would have to be like this:
```go
// Note: math.MaxInt was added in Go 1.17. We'd have to do something else for older versions.
if sz > C.sqlite3_int64(math.MaxInt) {
    return nil, fmt.Errorf(...)
}

cBuf := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
    Data: unsafe.Pointer(ptr),
    Len: int(sz),
    Cap: int(sz),
}))

res := make([]byte, int(sz))
copy(res, cBuf)
return res
```

sqlite3_opt_serialize.go
otoolep: What do you suggest as the most efficient way to copy the memory from C to the Go slice? Use memcpy under the covers, something like? https://gist.github.com/3d0c/d8583aa8ac8fbc43585620a20a8b9ac9

otoolep: That comment in mistaken. I meant copying in the other direction.

Do you suggest using memcpy to copy data from the Go slice to the memory address allocated by `sqlite3_malloc`?

otoolep: As per https://github.com/mattn/go-sqlite3/pull/1089#discussion_r969670006

sqlite3_opt_serialize.go
rittneje: If they aren't using libsqlite3, then these functions are always included (there is no opt-out).
If they are using libsqlite3, then they must opt-in via a build tag (`sqlite_serialize`) to preserve backwards compatibility in all cases.

Thus this condition should be `// +build !libsqlite3 sqlite_serialize`.

otoolep:
If you want me to squash the commits, I can do that. Just let me know.

rittneje:
> If you want me to squash the commits, I can do that. Just let me know.

Don't worry, we can squash during the merge.

.github/workflows/go.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: go-acc . -- -race -v -tags ""
 
       - name: 'Tags: libsqlite3'
-        run: go-acc . -- -race -v -tags "libsqlite3"
+        run: go-acc . -- -race -v -tags "libsqlite3 sqlite_omit_deserialize"
```
rittneje, line 44: These changes can be reverted.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"fmt"
+	"math"
+	"reflect"
+	"unsafe"
+)
+
+const (
+	SQLITEDeserializeFreeOnClose = 1
```
rittneje, line 24: Why do you need this? Can't you use the value from the sqlite3.h header file?

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+// Serialize returns a byte slice that is a serialization of the database.
+//
+// See https://www.sqlite.org/c3ref/serialize.html
+func (c *SQLiteConn) Serialize(schema string) ([]byte, error) {
```
rittneje, line 30: @otoolep What has been your typical use case for this method? I am wondering if it would be useful to have a `SerializeTo` version that takes an `io.Writer` instead of returning `[]byte`. That way if your intent is to stream it somewhere (like a file or HTTP response) we don't have to have two copies in RAM simultaneously.

It could be nice to have something similar for `Deserialize` with `io.Reader`, but it would be more complicated because of the buffering that is required.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,88 @@
...
+	}))
+
+	res := make([]byte, int(sz))
+	copy(res, cBuf)
+	return res, nil
+}
+
+// Deserialize causes the connection to disconnect from the current database
+// and then re-open as an in-memory database based on the contents of the
+// byte slice. If deserelization fails, error will contain the return code
```
rittneje, line 62:
```suggestion
// byte slice. If deserialization fails, error will contain the return code
```
Also, as written, there is no way for the caller to actually get the code without parsing it. Probably we should attempt to return an instance of `sqlite3.Error` if we intend for the caller to be able to examine it. Otherwise we should not mention this detail in the contract.

sqlite3_opt_serialize_omit.go
```diff
@@ -0,0 +1,21 @@
+// +build libsqlite3
```
rittneje, line 1:
```suggestion
// +build libsqlite3,!sqlite_serialize
```

sqlite3_opt_serialize.go
otoolep: I am not super familiar with cgo programming, so missed I can do that -- fixed.

sqlite3_opt_serialize.go
otoolep: Good idea, I would very much like to reduce memory usage. I've never been happy with the amount of memory copying going on in callers.

rqlite uses this method to snapshot the SQLite database to disk, and part of a process know as "Log Truncation". 

Serialization is triggered here: https://github.com/rqlite/rqlite/blob/697782280fc22ccf66c23693bb0a235538664257/store/store.go#L1347

The byte slice is then written to disk here: https://github.com/rqlite/rqlite/blob/697782280fc22ccf66c23693bb0a235538664257/store/store.go#L1394

The only detail is that the byte slice is first compressed before writing it to disk. I still think we could probably avoid the double copies.

otoolep: How about we get this PR in, and I follow up with a `SerializeTo()` call?

otoolep: That said, rqlite needs a copy of the SQLite database at a very specific instant, and that copy is not written immediately to disk (because the rqlite doesn't know the file path at the time the Snapshotting process asks for a copy of the SQLite database). So getting a copy of the database, and writing it to disk, happen at different times, possibly from different Goroutines.

So rqlite may not make immediately use of any `SerializeTo` but I'm going to think about it. I might be able to make it work.

rittneje: > How about we get this PR in, and I follow up with a SerializeTo() call?

Yes that sounds reasonable.

otoolep:
I think we're good now. Let me know there are more required changes.

otoolep:
What do you guys think, ready to merge?

mattn (approved):

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,84 @@
...
+	defer C.free(unsafe.Pointer(zSchema))
+
+	var sz C.sqlite3_int64
+	ptr := C.sqlite3_serialize(c.db, zSchema, &sz, 0)
+	if ptr == nil {
+		return nil, fmt.Errorf("serialize failed")
+	}
+	defer C.sqlite3_free(unsafe.Pointer(ptr))
+
+	if sz > C.sqlite3_int64(math.MaxInt) {
```
rittneje, line 41: @mattn FYI this constant was introduced in Go 1.17. The GitHub actions suggest that is the minimum required version for this library, but the go.mod file still says 1.16. Just wanted to confirm you have no concerns about this.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,84 @@
...
+	}))
+
+	res := make([]byte, int(sz))
+	copy(res, cBuf)
+	return res, nil
+}
+
+// Deserialize causes the connection to disconnect from the current database
+// and then re-open as an in-memory database based on the contents of the
+// byte slice. If deserialization fails, error will contain the return code
```
rittneje, line 58: As I mentioned, we don't want clients attempting to parse the error string. So either we should not mention any error code in the docs (and just say an error will be returned), or we should return an instance of `*sqlite3.Error` so they can get it in the normal way.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,84 @@
...
+	cBuf := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
+		Data: uintptr(unsafe.Pointer(tmpBuf)),
+		Len:  len(b),
+		Cap:  len(b),
+	}))
+	copy(cBuf, b)
+
+	rc := C.sqlite3_deserialize(c.db, zSchema, tmpBuf, C.sqlite3_int64(len(b)),
+		C.sqlite3_int64(len(b)), C.SQLITE_DESERIALIZE_FREEONCLOSE)
+	if rc != 0 {
```
rittneje, line 80:
```suggestion
	if rc != C.SQLITE_OK {
```

(they mean the same thing but I think using the error code constants is more clear)

sqlite3_opt_serialize_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"database/sql"
+	"database/sql/driver"
+	"os"
+	"testing"
+)
+
+func TestSerialize(t *testing.T) {
+	d := SQLiteDriver{}
+
+	srcConn, err := d.Open(":memory:")
```
rittneje, line 15: If you just use `:memory:` you can run into weird issues due to Go's connection pooling. Please use file:/TestSerialize?vfs=memdb` instead.

sqlite3_opt_serialize_test.go
```diff
@@ -0,0 +1,120 @@
...
+package sqlite3
+
+import (
+	"database/sql"
+	"database/sql/driver"
+	"os"
+	"testing"
+)
+
+func TestSerialize(t *testing.T) {
```
rittneje, line 12: Can we write this test to go through `database/sql` instead? That will also serve as a good example for developers for how to use it from their code.

sqlite3_opt_serialize_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Fatal("failed to create table:", err)
+	}
+	_, err = sqlite3conn.Exec(`INSERT INTO foo(name) VALUES("alice")`, nil)
+	if err != nil {
+		t.Fatal("failed to insert record:", err)
+	}
+	conn.Close()
+
+	// Read database file bytes from disk.
+	b, err := os.ReadFile(tempFilename)
```
rittneje, line 88: It should be noted that this is not particularly safe in production code as it is dependent upon the journal mode. (For example, if you are using WAL, then the resulting deserialized database may be incomplete or possibly corrupted.) In practice this particular use case (deserializing from a file) is better served by the backup API.

otoolep: Sure, happy to write a single test. But let offer a couple of comments first, see what you think.

This is a test, where the test clearly controls the environment 100%. This explicitly doesn't use the WAL (which would be problematic, as you say), there is no multi-process access involved, etc. So there is no chance of it failing due to that since the test code controls everything. While everything you say is true, none of it applies here. I consider this test robust.

Also, a test that uses Serialize/Deserialize in the same test makes me nervous. Imagine there is a bug in some shared code that both functions use. And the bug is hidden when you use both functions back-to-back (one function introduces the bug, the other reverses it). *That* is why I wrote the test this way -- call one of the functions and then *independently* verify it is doing the right thing.

If you are concerned that the existence of this could make people use the new functions incorrectly -- that the tests act as a form of documentation -- then that is a good point. I think it's a strong argument about rewriting the test as you say.

sqlite3_opt_serialize.go
otoolep: Done.

sqlite3_opt_serialize_test.go
otoolep: Sounds reasonable, let me take a look.

sqlite3_opt_serialize_test.go
rittneje: Yes, my concern is that someone might reference the test case as an example of how to use the feature.

Definitely I think having the round-trip test is important in order to test the actual feature. If you want to additionally include tests for serialize/deserialize in isolation, that can be okay, but you should add clear comments that doing that in production code can be unsafe for several reasons, and people should use the `Backup` method instead if they need to serialize to or deserialize from another database file.

sqlite3_opt_serialize.go
mattn: Is this constant an operating system cpu architecture dependent?

rittneje: The value is `GOARCH`-dependent, but it is always defined (in 1.17+).

sqlite3_opt_serialize_test.go
rittneje: @otoolep It doesn't look like this was done.

otoolep: Oh yes,  you're right -- I had forgotten this is on me. Thanks for reminding me.

otoolep: I rewrote the Serialization test. Do you want to take a look? Is this what you had in mind?

sqlite3_opt_serialize_test.go
otoolep: New test added.

sqlite3_opt_serialize_test.go
otoolep: New test added.

sqlite3_opt_serialize.go
otoolep: Fixed in the code.

sqlite3_opt_serialize.go
otoolep: Fixed in code.

sqlite3_opt_serialize.go
otoolep: @mattn has approved this PR, so I'll assume this conversation is resolved.

sqlite3_opt_serialize.go
```diff
@@ -0,0 +1,82 @@
...
+		Data: uintptr(unsafe.Pointer(tmpBuf)),
+		Len:  len(b),
+		Cap:  len(b),
+	}))
+	copy(cBuf, b)
+
+	rc := C.sqlite3_deserialize(c.db, zSchema, tmpBuf, C.sqlite3_int64(len(b)),
+		C.sqlite3_int64(len(b)), C.SQLITE_DESERIALIZE_FREEONCLOSE)
+	if rc != C.SQLITE_OK {
+		return fmt.Errorf("deserialize failed with return %v", rc)
```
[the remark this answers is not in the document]
otoolep: I'm not sure what you mean by "works". It compiles, and the tests pass, but I'm not sure I want to rely on something that is "not clear" in the docs.

```
git diff
diff --git a/sqlite3_opt_serialize.go b/sqlite3_opt_serialize.go
index 2560c43..e757463 100644
--- a/sqlite3_opt_serialize.go
+++ b/sqlite3_opt_serialize.go
@@ -1,3 +1,4 @@
 // +build !libsqlite3 sqlite_serialize
 
 package sqlite3
@@ -73,10 +74,7 @@ func (c *SQLiteConn) Deserialize(b []byte, schema string) error {
        }))
        copy(cBuf, b)
 
-       rc := C.sqlite3_deserialize(c.db, zSchema, tmpBuf, C.sqlite3_int64(len(b)),
+       C.sqlite3_deserialize(c.db, zSchema, tmpBuf, C.sqlite3_int64(len(b)),
                C.sqlite3_int64(len(b)), C.SQLITE_DESERIALIZE_FREEONCLOSE)
-       if rc != C.SQLITE_OK {
-               return fmt.Errorf("deserialize failed with return %v", rc)
-       }
-       return nil
+       return c.lastError()
 }
```

Does that answer your question?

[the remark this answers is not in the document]
rittneje: I was mostly just curious for future purposes, if we ever want this to return `*sqlite3.Error` type. The docs don't explicitly say whether `sqlite3_errmsg`, etc. will yield expected values if `sqlite3_deserialize` fails.

sqlite3_opt_serialize_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"testing"
+)
+
+func TestSerializeDeserialize(t *testing.T) {
+	// The driver's connection will be needed in order to serialization and deserialization
+	driverName := "TestSerializeDeserialize"
+	driverConns := []*SQLiteConn{}
+	sql.Register(driverName, &SQLiteDriver{
+		ConnectHook: func(conn *SQLiteConn) error {
+			driverConns = append(driverConns, conn)
```
rittneje, line 17: As I mentioned, this is not a safe way to do this. In practice, developers should use the `Raw` method like so:
```go
srcDb, err := sql.Open("sqlite3", srcTempFilename)
if err != nil {
    t.Fatal(err)
}
defer srcDb.Close()

// insert into src
...

srcConn, err := srcDb.Conn(context.Background())
if err != nil {
    t.Fatal(err)
}

var serialized []byte
if err := srcConn.Raw(func(raw interface{}) error {
    var err error
    serialized, err = raw.(*sqlite3.SQLiteConn).Serialize("")
    return err
}); err != nil {
    t.Fatal(err)
}

srcConn.Close()

destDb, err := sql.Open("sqlite3", destTempFilename)
if err != nil {
    t.Fatal(err)
}
defer destDb.Close()

destConn, err := destDb.Conn(context.Background())
if err != nil {
    t.Fatal(err)
}

if err := destConn.Raw(func(raw interface{}) error {
    return raw.(*sqlite3.SQLiteConn).Deserialize(serialized, "")
}); err != nil {
    t.Fatal(err)
}

destConn.Close()

// now query destDb to verify all the data is present
...
```

otoolep:
@rittneje -- I think we're good to merge this? Anything else need doing? The test has been updated, and all tests pass.

With this merged I can return to using this repo, and dump my fork.

rittneje (approved):

otoolep:
Thank you! Looking forward to moving back to this repo, and away from my fork.

mattn:
@otoolep Thank you.

PR #1110, opened 2022-11-17
solve the problem of aggregation query losing precision.

786744873:
when i use sql tools to execute sql。not lose precision
image
when i use sql tools to execute sql。not lose precision
image

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/mattn/go-sqlite3
+module github.com/786744873/go-sqlite3
 
 go 1.16
 
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -2223,7 +2223,10 @@ func (rc *SQLiteRows) nextSyncLocked(dest []driver.Value) error {
 				dest[i] = val
 			}
 		case C.SQLITE_FLOAT:
-			dest[i] = float64(C.sqlite3_column_double(rc.s.s, C.int(i)))
+			n := int(C.sqlite3_column_bytes(rc.s.s, C.int(i)))
+			s := C.GoStringN((*C.char)(unsafe.Pointer(C.sqlite3_column_text(rc.s.s, C.int(i)))), C.int(n))
+			f, _ := strconv.ParseFloat(s,64)
+			dest[i] = f
 		case C.SQLITE_BLOB:
 			p := C.sqlite3_column_blob(rc.s.s, C.int(i))
 			if p == nil {


rittneje:
Do you have a reproducible example of this issue? Because from my testing I think it is the other way around - if `25095.505999999998` is converted to a string by SQLite then it turns it into `25095.506`. The GUI tool you are comparing too might just always use `sqlite3_column_text`, which is why it produces the wrong answer. You can observe the same issue with the standard sqlite3 CLI by running `SELECT 25095.505999999998;`.

786744873:
**Here's my code ,I wish to be a contributor**

```
package main

import (
	"database/sql"
	"fmt"
	"log"
	"os"
)

func main() {
	os.Remove("./foo.db")

	db, err := sql.Open("sqlite3", "./foo.db")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	sqlStmt := `
	create table foo (id integer not null primary key, value text);
	delete from foo;
	`
	_, err = db.Exec(sqlStmt)
	if err != nil {
		log.Printf("%q: %s\n", err, sqlStmt)
		return
	}

	tx, err := db.Begin()
	if err != nil {
		log.Fatal(err)
	}
	stmt, err := tx.Prepare("insert into foo(id, value) values(?, ?)")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()
	fmt.Println("mock data:")
	for i := 0; i < 10; i++ {
		_, err = stmt.Exec(i, "2509.5506")
		if err != nil {
			log.Fatal(err)
		}
	}
	err = tx.Commit()
	if err != nil {
		log.Fatal(err)
	}

	rows, err := db.Query("select id, value from foo")
	if err != nil {
		log.Fatal(err)
	}
	defer rows.Close()
	for rows.Next() {
		var id int
		var name string
		err = rows.Scan(&id, &name)
		if err != nil {
			log.Fatal(err)
		}
		fmt.Println(id, name)
	}
	err = rows.Err()
	if err != nil {
		log.Fatal(err)
	}

	var value string

	fmt.Println("sum error")
	stmt, err = db.Prepare("select sum(value) from foo")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()

	err = stmt.QueryRow().Scan(&value)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(value)

	fmt.Println("avg error")
	stmt, err = db.Prepare("select avg(value) from foo")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()
	err = stmt.QueryRow().Scan(&value)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(value)

	fmt.Println("max OK")
	stmt, err = db.Prepare("select max(value) from foo")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()
	err = stmt.QueryRow().Scan(&value)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(value)

	fmt.Println("min OK")
	stmt, err = db.Prepare("select min(value) from foo")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()
	err = stmt.QueryRow().Scan(&value)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(value)

	fmt.Println("count OK")
	stmt, err = db.Prepare("select count(value) from foo")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close()
	err = stmt.QueryRow().Scan(&value)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(value)

}
```

**this is execute result**
image

It's my honor to contribute code. thanks a lot

rittneje:
These sorts of issues are par for the course when working with floating point numbers. For example, Go produces the exact same results. https://go.dev/play/p/i1AWFLb7KSv

Thanks for the contribution, but the current code is actually working as expected.

786744873:
Did you run the example code I gave you?
This is a bug in your packaging of c-sqlite,beacause you use the wrong way! 
This cannot be blamed on the precision of the go language itself.
If the result is `25095.50599999998`,how can peolple to use?The user does not know that the result of the source is wrong！
Are you make a  joke?

image

image

rittneje:
This package is not using SQLite the wrong way. The issue you are experiencing has nothing to do with SQLite in particular, and everything to do with how IEEE standard floating point numbers work. The Go example I linked was meant to show exactly that. I suggest looking more into this topic if you are not familiar with it.

In other words, SQLite itself produces the correct value as per standard floating point arithmetic. This is the value that this library is returning verbatim.

The reason you are getting a different answer from other applications is because they are always converting the floating point result to a string. And there appears to be some peculiarity in SQLite itself here where this process turns `25095.505999999998` into `25095.506`. In your specific case you found that peculiarity beneficial as it happened to counteract the round-off issue inherent to floating point arithmetic.

If your application cannot tolerate the nuances of floating point arithmetic, it needs to be dealt with there. You may choose to perform some rounding explicitly, or leverage the aforementioned peculiarity by casting to string, or use fixed point, to name a few possibilities.

786744873:
What I need to clarify is i am not converting the floating point result to a string. 
I was only try get this `select sum(value)` from db which really value is `25095.506`. 
The source code here turns the data into `25095.505999999998`. In fact, the data stored in the database is `25095.506`.
[https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L1960](https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L1960)

Have you carefully looked at the table structure in the sample code I gave?
```
create table foo (id integer not null primary key, value text)
```
the type of value is `text`.Sqlite enable make a sum calucation, if this `text` value is type of number

786744873:
[foo.tar.gz](https://github.com/mattn/go-sqlite3/files/10038987/foo.tar.gz)
This is the database file.you can make a query by your tools.
```
select sum(value) from foo
```

rittneje:
> the type of value is text

SQLite is flexibly typed, meaning that the declared column type is irrelevant (unless you use strict tables). What actually matters is the type of each individual value. In this case, those are also strings because you wrote `stmt.Exec(i, "2509.5506")` as opposed to `stmt.Exec(i, 2509.5506)`. See https://sqlite.org/datatype3.html for more information.

That aside, because you are calculating `SUM`, each of those values must be converted from string to floating point, because that is how SQLite does it. And the actual result of `SUM` is a floating point number. See https://www.sqlite.org/lang_aggfunc.html#sumunc
> If any input to sum() is neither an integer nor a NULL, then sum() returns a floating point value which is an approximation of the mathematical sum.

 As I demonstrated, any application that does standard floating point arithmetic to add `25095.506` 10 times is going to get `25095.505999999998` as a result.

If you convert this value to a string via `sqlite3_column_text`, it happens to produce the value `"25095.506"`. I am not really sure why, but my guess is it is due to some implicit rounding. Thus the reason that whatever graphical tool you are using is giving you `25095.506` is because that tool likely is just calling `sqlite3_column_text` regardless of what SQLite says the type of the value is.

PR #1125, opened 2023-01-26
Document that "Z" is trimmed when parsing time

emersion:


rittneje:
@mattn The current behavior is actually pretty strange. Right now this library doesn't even support timezones (other than Z) for any format other than the first two.

Probably we should remove the part where it chops off Z and fix the list of formats to:
```go
var SQLiteTimestampFormats = []string{
	"2006-01-02 15:04:05.999999999Z07:00",
	"2006-01-02 15:04:05.999999999",
	"2006-01-02T15:04:05.999999999Z07:00",
	"2006-01-02T15:04:05.999999999",
	"2006-01-02 15:04:05Z07:00",
	"2006-01-02 15:04:05",
	"2006-01-02T15:04:05Z07:00",
	"2006-01-02T15:04:05",
	"2006-01-02 15:04Z07:00",
	"2006-01-02 15:04",
	"2006-01-02T15:04Z07:00",
	"2006-01-02T15:04",
	"2006-01-02Z07:00",
	"2006-01-02",
}
```

However, that could get into some interesting questions wrt backwards compatibility.

It would also be worth mentioning that if the parsed timestamp does not include time zone information it will assume UTC (and not local).

PR #1128, opened 2023-02-02
Replace namedValue with driver.NamedValue to avoid copying exec/query args

charlievieth:
This commit replaces the `namedValue` type with `driver.NamedValue` this eliminates the need to create copy of the arguments slice in `{SQLiteConn,SQLiteStmt}.{QueryContext,ExecContext}`.  

The `driver.NamedValue` type was added in 2016 (https://github.com/golang/go/commit/707a83341b8c7973f4e0fce731fa279c618f233b) so there shouldn't be any compatibility issues.


rittneje:
LGTM. (`driver.NamedValue` was added in Go 1.8 specifically.) @mattn any concerns?

mattn:
LGTM

PR #1134, opened 2023-02-09
Proposal: sub-packages to enable optional features

emersion:
A common pain point when using go-sqlite3 is that enabling optional features requires adding a build tag to all `go` command invocations. For instance, when a program depends on FTS5, it's no longer possible to run `go build`, `go test`, etc as usual. End users need to explicitly add `-tags sqlite_fts5` every time.

Here is a proposal to improve this. A new `fts5` sub-package is introduce. Any program which depends on FTS5 can import it to turn on FTS5:

```go
import _ "github.com/mattn/go-sqlite3/fts5"
```

What do you think?

### DIFF
--- a/fts5/fts5.go
+++ b/fts5/fts5.go
@@ -1,0 +1,12 @@
+// Copyright (C) 2023 Simon Ser <contact@emersion.fr>
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file.
+
+package fts5
+
+/*
+#cgo CFLAGS: -DSQLITE_ENABLE_FTS5
+#cgo LDFLAGS: -lm
+*/
+import "C"
--- a/sqlite3_opt_fts5.go
+++ b/sqlite3_opt_fts5.go
@@ -7,8 +7,4 @@
 
 package sqlite3
 
-/*
-#cgo CFLAGS: -DSQLITE_ENABLE_FTS5
-#cgo LDFLAGS: -lm
-*/
-import "C"
+import _ "github.com/mattn/go-sqlite3/fts5"


emersion:
Yes, I have tested this and it works correctly.

AFAIU the cgo flags are *not* package-scoped, they are module-scoped. See: https://github.com/golang/go/issues/40041

rittneje:
I just ran a simplified test case. From what I can tell, the `#cgo` directives are indeed package-scoped, not module-scoped.

cgotest/go.mod
```go
module cgotest

go 1.18
```

cgotest/main.go
```go
package main

import _ "cgotest/subpackage"

/*
#include <sqlite3.h>
*/
import "C"

func main() {
	C.sqlite3_open_v2(nil, nil, 0, nil)
}
```

cgotest/subpackage/c.go
```go
package subpackage

/*
#cgo LDFLAGS: -lsqlite3
*/
import "C"
```

```
$ go run main.go 
# command-line-arguments
Undefined symbols for architecture x86_64:
  "_sqlite3_open_v2", referenced from:
      __cgo_6884829902c1_Cfunc_sqlite3_open_v2 in _x002.o
     (maybe you meant: __cgo_6884829902c1_Cfunc_sqlite3_open_v2)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

emersion:
On my system this builds just fine (and fails at runtime because of the `nil` parameters):

```
> go run main.go
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7f21b6f2da2d]
```

rittneje:
Okay, based on further testing, it seems this behavior changed in Go 1.19. However, it was not mentioned in the release notes. And I'm not sure if it was intentional, since the cgo docs make no mention of modules or sub-packages. I'm going to file an issue with the Go authors for this.

In the meantime, I don't think we can accept this change, since this module intends to support Go 1.18.

emersion:
Note, the result of this discussion is at https://github.com/golang/go/issues/58438#issuecomment-1424957107.

emersion:
In the end I opted to bundle FTS5 in a separate Go package: https://git.sr.ht/~emersion/go-sqlite3-fts5

emersion:
Now that this package [depends on Go 1.19](https://github.com/mattn/go-sqlite3/commit/c91bca4fb499831d1bf62fb2b3019338d5a3eb41), would you be open to considering this PR for inclusion?

PR #1141, opened 2023-03-10
go.mod: go 1.18

tendervittles:
go.mod's go version should match the oldest entry in github workflow.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,5 @@
 module github.com/mattn/go-sqlite3
 
-go 1.16
+go 1.18
 
-retract (
- [v2.0.0+incompatible, v2.0.6+incompatible] // Accidental; no major changes or features.
-)
+retract [v2.0.0+incompatible, v2.0.6+incompatible] // Accidental; no major changes or features.


rittneje:
@mattn any concerns?

RaphaelKosik6z:
It's generally a good practice to ensure that the version of Go specified in the go.mod file matches the oldest version of Go that's supported by your GitHub workflow. This can help ensure compatibility and prevent potential issues from arising due to version mismatches. It's also important to keep both the go.mod file and the GitHub workflow up-to-date as new versions of Go are released, to ensure that your code remains compatible with the latest tools and libraries.

OswaldoBeville9Bw:
The statement you provided appears to be a guideline for maintaining consistency between the version of Go specified in a project's go.mod file and the version of Go used in the project's GitHub workflow configuration.

This guideline suggests that the go version specified in go.mod should match the oldest version of Go used in the project's GitHub workflow. This helps ensure that the project's dependencies are compatible and that the code can be compiled and tested using the specified version of Go.

Maintaining consistency between the go.mod file and the GitHub workflow configuration is important for ensuring the reliability and reproducibility of a project's builds and tests.

WilfordWetzelj:
It seems like you have provided a guideline for maintaining consistency between the version of Go specified in a project's go.mod file and the version of Go used in the project's GitHub workflow configuration.

This guideline recommends that the go version specified in go.mod should match the oldest version of Go used in the project's GitHub workflow. By doing so, the project's dependencies are kept compatible and the code can be compiled and tested using the specified version of Go.

This is an important step for ensuring that a project's builds and tests are reliable and reproducible. Consistency between the go.mod file and the GitHub workflow configuration helps avoid potential compatibility issues that can arise from using different versions of Go in different parts of the project.

dolmen:
The right fix is not to upgrade the version in go.mod but instead to extend the coverage with older versions of Go in the CI.

PR #1144, opened 2023-03-23
Support "BEGIN CONCURRENT"

bprosnitz:
SQLite supports a "BEGIN CONCURRENT" transaction mode which cannot currently be used with the go-sqlite3 library.

See https://www.sqlite.org/cgi/src/doc/begin-concurrent/doc/begin_concurrent.md for details


rittneje:
This is not a feature of standard SQLite. I think it requires a custom build?

PR #1149, opened 2023-04-05
Fix virtual table example.

andrzh:
`csts` is used on 66, but parameter passed in is `cst` leading to a compile error.


rittneje:
Thanks!

PR #1156, opened 2023-04-21
preupdate_hook: move convertAssign code to preupdate_hook file

charlievieth:
Move the convertAssign logic from convert.go to
sqlite3_opt_preupdate_hook.go since that is the
only place it is used.


dolmen (commented):
What about just adding the build tags to convert.go?

```
//go:build sqlite_preupdate_hook
// +build sqlite_preupdate_hook
```

charlievieth:
> What about just adding the build tags to convert.go?
> 
> ```
> //go:build sqlite_preupdate_hook
> // +build sqlite_preupdate_hook
> ```

That's also a valid solution. My motivation for moving it to sqlite3_opt_preupdate_hook.go is that it is a relatively small file, it's nice to consolidate logic, and most of the file names match their respective build tags so I felt that it would be clearer / more consistent to move it there.

charlievieth:
@dolmen could you take another look at this PR?

PR #1157, opened 2023-04-21
sqlite3: save an alloc during bind by using time.Time.AppendFormat

charlievieth:
This commit changes bind() to use time.Time.AppendFormat, which returns a byte slice, instead of time.Time.Format, which returns a string that must then be converted to a byte slice. This should save an allocation.


dolmen:
Note: [`time.Time.AppendFormat`](https://pkg.go.dev/time#Time.AppendFormat) has been added with Go 1.5.

charlievieth:
@dolmen Any chance you could take another look at this PR?

PR #1166, opened 2023-05-31
Update amalgamation code

mattn:


otoolep:
Thanks for the upgrade to SQLite 3.42.0!

PR #1167, opened 2023-06-01
New txlock options for hinting RO transaction

hifi:
If an application doesn't have a retry logic when a read only transaction is being automatically upgraded to a write one in the default deferred mode SQLite will immediately return SQLITE_BUSY instead of honoring the busy timeout.

To workaround this one can use _txlock=immediate option to always go into write lock when starting a transaction. Depending on the application this might not be always wanted.

The Go SQL layer provides a mechanism to request a read only transaction but SQLite doesn't support completely disabling writes within a transaction. However, we can use it to hint that we might not do any writes so we'd instead want a deferred BEGIN.

Because this isn't very standards compliant in any way adding new _txlock options was chosen so it can be selected when desired.


hifi:
Rethinking this since SQLite supports `PRAGMA query_only`, the correct way to implement this is to use separate connection pool for read only transactions as the PRAGMA will guarantee no writes will happen.

We implemented this approach downstream so our main connection pool has `_txlock=immediate` and the read only pool doesn't but has `_query_only=true`. The only problem with the approach is we had to use a wrapper to pick between the pools transparently.

Issue #1164, opened 2023-05-17
Build failure with musl 1.2.4


jefferyto:
OpenWrt updated musl to 1.2.4 and we are seeing this build failure with [ooni/probe-cli](https://github.com/ooni/probe-cli) (which uses go-sqlite3):

```
sqlite3-binding.c:35911:42: error: 'pread64' undeclared here (not in a function); did you mean 'pread'?
35911 |   { "pread64",      (sqlite3_syscall_ptr)pread64,    0  },
      |                                          ^~~~~~~
      |                                          pread
sqlite3-binding.c:35929:42: error: 'pwrite64' undeclared here (not in a function); did you mean 'pwrite'?
35929 |   { "pwrite64",     (sqlite3_syscall_ptr)pwrite64,   0  },
      |                                          ^~~~~~~~
      |                                          pwrite
sqlite3-binding.c: In function 'seekAndRead':
sqlite3-binding.c:35915:49: error: unknown type name 'off64_t'; did you mean 'off_t'?
35915 | #define osPread64 ((ssize_t(*)(int,void*,size_t,off64_t))aSyscall[10].pCurrent)
      |                                                 ^~~~~~~
sqlite3-binding.c:38777:11: note: in expansion of macro 'osPread64'
38777 |     got = osPread64(id->h, pBuf, cnt, offset);
      |           ^~~~~~~~~
sqlite3-binding.c:35915:58: error: expected ')' before 'aSyscall'
35915 | #define osPread64 ((ssize_t(*)(int,void*,size_t,off64_t))aSyscall[10].pCurrent)
      |                   ~                                      ^~~~~~~~
sqlite3-binding.c:38777:11: note: in expansion of macro 'osPread64'
38777 |     got = osPread64(id->h, pBuf, cnt, offset);
      |           ^~~~~~~~~
sqlite3-binding.c: In function 'seekAndWriteFd':
sqlite3-binding.c:35933:57: error: unknown type name 'off64_t'; did you mean 'off_t'?
35933 | #define osPwrite64  ((ssize_t(*)(int,const void*,size_t,off64_t))\
      |                                                         ^~~~~~~
sqlite3-binding.c:38906:17: note: in expansion of macro 'osPwrite64'
38906 |   do{ rc = (int)osPwrite64(fd, pBuf, nBuf, iOff);}while( rc<0 && errno==EINTR);
      |                 ^~~~~~~~~~
[...]

PR #1177, opened 2023-07-10
Fix musl build (#1164)

leso-kn:
As pointed out by @jefferyto the LFS64 interfaces were marked as deprecated in musl 1.2.4 ([release notes](https://musl.libc.org/releases.html), [commit](https://git.musl-libc.org/cgit/musl/commit/?h=v1.2.4&id=25e6fee27f4a293728dd15b659170e7b9c7db9bc)), thus go-sqlite3 currently fails to compile on musl.

This PR adjusts CLFAGS to let the C implementation use `pread()`/`pwrite()` instead of `pread64()`/`pwrite64()`.

Note: This change does not affect existing glibc builds as suggested by the [pread64 man page](https://linux.die.net/man/2/pread64):
> On Linux, the underlying system calls were renamed in kernel 2.6: pread() became pread64(), and pwrite() became pwrite64(). The system call numbers remained the same. The glibc pread() and pwrite() wrapper functions transparently deal with the change.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -21,7 +21,6 @@ package sqlite3
 #cgo CFLAGS: -DSQLITE_DEFAULT_WAL_SYNCHRONOUS=1
 #cgo CFLAGS: -DSQLITE_ENABLE_UPDATE_DELETE_LIMIT
 #cgo CFLAGS: -Wno-deprecated-declarations
-#cgo linux,!android CFLAGS: -DHAVE_PREAD64=1 -DHAVE_PWRITE64=1
 #cgo openbsd CFLAGS: -I/usr/local/include
 #cgo openbsd LDFLAGS: -L/usr/local/lib
 #ifndef USE_LIBSQLITE3
@@ -46,6 +45,18 @@ package sqlite3
 
 #ifndef SQLITE_DETERMINISTIC
 # define SQLITE_DETERMINISTIC 0
+#endif
+
+#if defined(HAVE_PREAD64) && defined(HAVE_PWRITE64)
+# undef USE_PREAD
+# undef USE_PWRITE
+# define USE_PREAD64 1
+# define USE_PWRITE64 1
+#elif defined(HAVE_PREAD) && defined(HAVE_PWRITE)
+# undef USE_PREAD
+# undef USE_PWRITE
+# define USE_PREAD64 1
+# define USE_PWRITE64 1
 #endif
 
 static int


omerxx:
Honestly unsure how this isn't urgent to users out there but it saved me.
So for what it's worth I added this to my `go.mod`:
```golang
replace github.com/mattn/go-sqlite3 => github.com/leso-kn/go-sqlite3 v0.0.0-20230710125852-03158dc838ed
```

ptrcnull:
hey @mattn could we get this merged? it's a really small non-breaking change that would fix all musl builds

mattn:
Can you add test for the musl build ?

rittneje:
I had to do some digging here to check whether this change was safe, because on the surface it isn't. `pread` takes an `off_t` while `pread64` takes an `off64_t`. So in a vacuum, this would break on systems where `off_t` is not 64 bits (overflow). (Note that the comment in the man page is referring to how the glibc wrapper invokes the underlying syscalls and is not relevant to this discussion.)

However, SQLite itself also sets `_FILE_OFFSET_BITS` to 64 (unless you set `SQLITE_DISABLE_LFS`), which forces `off_t` to be an alias for `off64_t`.

I do think there is a bug in the C library though, since it doesn't appear SQLite will attempt to constrain itself to 32 bits if LFS is disabled.

ruzv:
any progress update on this? 
musl 1.2.4 was released in May 1, 2023

leso-kn:
Thanks for the feedback @mattn, I've changed the patch to your version!

mattn:
>Thanks for the feedback @mattn, I've changed the patch to your version!

Sorry, I made you confused. I just said that we can merge previous version of those changes.

mattn:
BTW, now I could reproduce this in https://github.com/mattn/nostr-relay/actions/runs/7151889396/job/19476647857

and confirmed with 

```
#cgo linux,!android CFLAGS: -DHAVE_PREAD=1 -DHAVE_PWRITE=1 -DHAVE_PREAD64=1 -DHAVE_PWRITE64=1
```

https://github.com/mattn/nostr-relay/pull/2/files#diff-89278323f4af3d0fc58cd9680b5fa55549b914cb147c5f69b9cc4d06d26a780bR24

mattn:
Or we might be possible to remove this line. (not make sure)

james-d-elliott:
For reference if anyone is having this issue in their CI/CD pipelines this affects the latest published golang alpine container image `golang:1.21.5-alpine` but doesn't affect `golang:1.21.4-alpine`. I suspect the 1.21.5 image was republished with the latest alpine/musl. It's also fairly likely that it'll affect the latest patch of each supported go version such as `1.20.12` as they likely republish these when the base image changes.

mattn:
Looks good to me.

I checked this change in https://github.com/mattn/nostr-relay/pull/2

mattn:
Hmm, still https://github.com/mattn/go-sqlite3/actions/runs/7143115977/job/19601740234?pr=1177 cc @leso-kn

itizir:
> Hmm, still https://github.com/mattn/go-sqlite3/actions/runs/7143115977/job/19601740234?pr=1177 cc @leso-kn

The job failure is unlikely related to the changes in this PR, and is instead due to changes in the latest amalgamation update: failed the same way here https://github.com/mattn/go-sqlite3/actions/runs/7195332329/job/19597759637?pr=802

And if you check what ran on latest release (v1.14.18), it was actually running with the previous tag (v1.14.17) checked out: https://github.com/mattn/go-sqlite3/actions/runs/6755304725/job/18363622922

leso-kn:
Hmm, okay considering the job failure perhaps my testing through Alpine's CI didn't cover some cases, though this seems strange. I suggest reverting the changes to the initial submitted version

leso-kn:
@mattn Could you trigger the workflow with these changes?

itizir:
> @mattn Could you trigger the workflow with these changes?

I think both your 55b17a8 and 65d6085 would fix the build. The issue is that the `_example/simple/Dockerfile` would need to have the locally checked out `go-sqlite3` available to the docker build, and do an appropriate `go mod edit -replace` to use that instead...

leso-kn:
In that case, I have either versions available on branches[[1]](https://github.com/leso-kn/go-sqlite3/tree/pread-pwrite-musl)[[2]](https://github.com/leso-kn/go-sqlite3/tree/pread-pwrite-musl-alternative) :) Let me know which one's preferred and I'll provide that.

itizir:
Did a quick fix here: https://github.com/mattn/go-sqlite3/pull/1201
Let me know if it makes sense.

itizir:
Right! If you now rebase, the pipeline should pass. And maybe you can even also revert [d9e2789](https://github.com/mattn/go-sqlite3/commit/d9e2789502bf576eabd0232f4fad9bd4abcd4f31)...

leso-kn:
Rebased and switched back to the `#ifdef`-variant

mattn:
LGTM, Thank you.

ruzv:
Is there an ETA for the next release that would include this fix?

james-d-elliott:
-7 hours ;) https://github.com/mattn/go-sqlite3/releases/tag/v1.14.19

Issue #968, opened 2021-08-12
Proposal for reading SQLite databases embedded in Go executables


GeekStocks:
# Proposal for reading SQLite databases embedded in Go executables

## Purpose
Numerous forums in **multiple** programming languages are littered with questions similar to this: 
> *"How can I read my preexisting SQLite database in memory mode? Why can't I just pass a memory pointer to my driver's open command?"*

The disappointing answer to these questions is simply that the authors of SQLite have not exposed a method to do so. Driver authors / maintainers like @mattn for this package cannot interface to something that doesn't exist. Given this, the "fallback" answer usually wins the day: open SQLite using `":memory:"` mode and then `Exec` a SQL script containing a backup "dump" of the database. Once the overhead of this loading process has occurred, the user has the environment they want.

But is there a better way that could avoid this loading overhead? **The purpose of this proposal is to explore and solicit comments for a proposed custom SQLite VFS module that would work in tandem with `mattn/go-sqlite3` and Go's new `embed` package.** You are encouraged to comment if you can add to the knowledge presented here and/or see another way to accomplish the goal.

## Trying a different approach
Most observed approaches to this problem start with the desired `:memory:` mode and quickly run into trouble with no pointer parameter available. This proposal avoids SQLite's memory mode altogether and seeks to use the *existing* SQLite URI mode to open an embedded database. By virtue of being embedded the database is already in memory and accessable via a variable, so opening it is all that remains. 

As of Go version 1.16 released in February, 2021 embedding a SQLite database into a Go executable is now trivial:

```
package main

import "embed"

//go:embed specimen.db
var embeddedDatabase embed.FS
```

Go's `embed` package offers `string`, `[]byte`, and the new `embed.FS` data types as embedded possibilities. Users of the "fallback" solution can now at least embed their SQL script backup "dumps" as a string. Since this author knows no method  to load SQLite databases from `[]byte`, the focus of this proposal is on the new `embed.FS` option shown above.

*It is important to note* that embedding digital assets into Go executables **is a read-only process**:

> An FS is a read-only collection of files, usually initialized with a //go:embed directive

This limits the usefulness of this proposal; if a project *must* have write ability on the SQLite database then the "fallback" method remains the only choice. But as the use cases for read-only databases are legion, we soldier on.

Under the hood SQLite uses a *standard* Virtual File System (VFS) to interpret the `open` string and locate files. The VFS chosen depends on the OS of course:

[...]

PR #1188, opened 2023-09-04
Simple read-only SQLite databases embedded in Go executables

dmarker:
The deserialize support that was added for #1089 doesn't quite address issue #968. But it comes very close. With this code you can basically follow along with a [SO from gizlu](https://stackoverflow.com/questions/68487191/embed-sqlite-database-into-c-executable). He even provided a [working C example](https://gist.github.com/gizlu/c6994b97b285a611e75df61482aa165c).

There is a limit due to SQLite only accepting one "mode=" in the DSN string. The C library would have to provide a Read-only memory option to have a named schema. Right now only "main" can be used, therefore only one embedded DB at a time can be used.


rittneje:
It is unclear to me whether this could run afoul of the rules of cgo. So it might be good to explicitly pin the memory first. Once you do that, it means this new method is not specific to embedded databases per se, but rather can be used as long as you guarantee that the byte slice in question will not be modified.

dmarker:
I'm relying on go GC never collecting a global since it is always reachable and I'm probably naive in believing it won't move it either. It seems it wouldn't so long as their value doesn't alter, but I will need to research that.

It has been a while since I've written go code. You can tell as I mention in the test I don't make the byte array const (you can't in go!).

I am certain you can only have a global for `embed`. The name DeserializeEmbedded is at least a hint that it really does in fact need to be embedded (actually global). I'll need to look more into runtime.Pinned to see if that will give the necessary guarantee. I would mostly be interested in sticking to the embed case so I don't have to worry about unpinning later.

rittneje:
> I'm probably naive in believing it won't move it either

The Go authors have reserved the right to implement a moving garbage collector in the future, which theoretically could affect global variables. Probably for the best to account for that now to avoid issues down the line.

rittneje:
You should be able to use built in build tags to only support the new method in 1.21 instead of updating go.mod.

dmarker:
I'll update to do that. Looks like the tag is for the whole file so I have some refactoring as well as retesting. Thank you for your suggestion!

dmarker:
Sorry about the delay in refactor, I'm in the middle of moving.

mikejsavage:
Any chance this could be merged?

PR #1190, opened 2023-09-30
Add SetErrorLog function

c4rlo:
This function sets a callback that SQLite invokes when it detects an anomaly. See https://sqlite.org/errlog.html.

## Summary by CodeRabbit

* **New Features**
  * Added an error logging callback API for receiving SQLite warnings and error messages.
  * Added constants for common SQLite notice and warning return codes.
  * Error callbacks now include the SQLite error code and descriptive message.

* **Bug Fixes**
  * Improved reporting of SQLite anomalies that previously could go unnoticed.

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -16,6 +16,7 @@ package sqlite3
 #else
 #include <sqlite3.h>
 #endif
+#include <stdint.h>
 #include <stdlib.h>
 
 void _sqlite3_result_text(sqlite3_context* ctx, const char* s, int n);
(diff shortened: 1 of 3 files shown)


c4rlo:
I've just rebased this.

Any word from the maintainers on this? I think this would be a good enhancement to have. Quoting the SQLite docs linked above:

> This mechanism is very helpful in tracking obscure problems that occur rarely and in the field. Application developers are encouraged to take advantage of the error logging facility of SQLite in their products, as it is very low CPU and memory cost but can be a huge aid for debugging.

PR #1195, opened 2023-10-23
bugfix: after executing statement table lock was not cleared

letFunny:
When we execute a statement, we step once and return. If the query returned more than one row, we would not have consumed everything so the underlying table would still be locked even when the execution had finished.

This change resets the statement after execution so that the locks can be cleared.

Apart from the included test, the following go program serves to reproduce the issue:
```golang
package main

import (
	"context"
	"database/sql"

)

func main() {
	ctx := context.Background()
	sqldb, err := sql.Open("sqlite3", "file:test.db?cache=shared&mode=memory")

	stmt, err := sqldb.PrepareContext(ctx, "CREATE TABLE person (id number, name text);")
	stmt.Exec()
	stmt, err = sqldb.PrepareContext(ctx, "INSERT INTO person VALUES (1, 'Fred'), (2, 'Susan');")
	stmt.Exec()

	stmt, err = sqldb.PrepareContext(ctx, "SELECT name FROM person;")
	_, err = stmt.ExecContext(context.Background())

	stmt, err = sqldb.Prepare("DROP TABLE person;")
	_, err = stmt.ExecContext(ctx)
	if err != nil {
		panic(err) // panic: database table is locked
	}
}
```

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -2083,6 +2083,12 @@ func (s *SQLiteStmt) execSync(args []driver.NamedValue) (driver.Result, error) {
 		return nil, err
 	}
 
+	rv = C.sqlite3_reset(s.s)
+	if rv != C.SQLITE_OK {
+		err := s.c.lastError()
+		C.sqlite3_clear_bindings(s.s)
+		return nil, err
+	}
 	return &SQLiteResult{id: int64(rowid), changes: int64(changes)}, nil
 }
 
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -2101,6 +2101,7 @@ var tests = []testing.InternalTest{
 	{Name: "TestManyQueryRow", F: testManyQueryRow},
 	{Name: "TestTxQuery", F: testTxQuery},
 	{Name: "TestPreparedStmt", F: testPreparedStmt},
+	{Name: "TestExecStmtDoesNotBlockTable", F: testExecStmtDoesNotBlockTable},
 	{Name: "TestExecEmptyQuery", F: testExecEmptyQuery},
 }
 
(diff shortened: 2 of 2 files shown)


sqlite3.go
```diff
@@ -2083,6 +2083,12 @@ func (s *SQLiteStmt) execSync(args []driver.NamedValue) (driver.Result, error) {
 		return nil, err
 	}
 
+	rv = C.sqlite3_reset(s.s)
+	if rv != C.SQLITE_OK {
```
letFunny, line 2087: Per the sqlite documentation ([ref](https://www.sqlite.org/c3ref/reset.html)),
> If the most recent call to [sqlite3_step(S)](https://www.sqlite.org/c3ref/step.html) for the [prepared statement](https://www.sqlite.org/c3ref/stmt.html) S indicated an error, then [sqlite3_reset(S)](https://www.sqlite.org/c3ref/reset.html) returns an appropriate [error code](https://www.sqlite.org/rescode.html). The [sqlite3_reset(S)](https://www.sqlite.org/c3ref/reset.html) interface might also return an [error code](https://www.sqlite.org/rescode.html) if there were no prior errors but the process of resetting the prepared statement caused a new error. For example, if an [INSERT](https://www.sqlite.org/lang_insert.html) statement with a [RETURNING](https://www.sqlite.org/lang_returning.html) clause is only stepped one time, that one call to [sqlite3_step(S)](https://www.sqlite.org/c3ref/step.html) might return SQLITE_ROW but the overall statement might still fail and the [sqlite3_reset(S)](https://www.sqlite.org/c3ref/reset.html) call might return SQLITE_BUSY if locking constraints prevent the database change from committing. Therefore, it is important that applications check the return code from [sqlite3_reset(S)](https://www.sqlite.org/c3ref/reset.html) even if no prior call to [sqlite3_step(S)](https://www.sqlite.org/c3ref/step.html) indicated a problem.

that is why we consider the execution failed if `reset` returns an error code.

letFunny:
@mattn Can you take a look to see if this PR makes sense? Should I create an issue along with it?

mattn:
As far as I look this changes, looks good to me. @rittneje could you please take a look?

letFunny:
@rittneje can you take a look?

PR #1200, opened 2023-11-23
Fix panic: sql: Register called twice for driver sqlite3

yjbdsky:
Fix panic: sql: Register called twice for driver sqlite3

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -239,9 +239,18 @@ const (
 var driverName = "sqlite3"
 
 func init() {
-	if driverName != "" {
+	if driverName != "" && !isRegistered(driverName) {
 		sql.Register(driverName, &SQLiteDriver{})
 	}
+}
+
+func isRegistered(name string) bool {
+	for _, v := range sql.Drivers() {
+		if v == name {
+			return true
+		}
+	}
+	return false
 }
 
 // Version returns SQLite library version information.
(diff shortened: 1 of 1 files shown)


mattn:
please explain reproduce steps

yjbdsky:
For some reason, my package and the dependency package both   import:  _ "github.com/mattn/go-sqlite3"

yjbdsky:
I just used sync.once instead of judging whether it has been registered through the array value.　please check

rittneje:
@yjbdsky One of two things is happening:
1. If it is caused by two different packages both importing this one, then it is a compiler bug, since a given package is not supposed to get initialized multiple times, no matter how many times it is imported. In this case, please file a bug against Go with a clear minimal reproducer.
2. It is caused by another package you are importing (directly or indirectly) also registering a driver under the name "sqlite3". In this case, you should either fix your code to only import one or the other, or get rid of the other registration.

Either way, I don't think this change should be accepted as it masks a real bug.

yjbdsky:
okay, i will close this pull

yjbdsky:
okay, i will close this pull

yjbdsky:
Both of my dependency packages use sqlite3, because the dependency level is deep. Is there an option to allow repeated registration when I know the error clearly?

rittneje:
Like I said, if the repeated registration is really caused by multiple imports of this package, then it is a compiler bug. If it is caused by another package also trying to register a driver named “sqlite3”, then either you need to fix that package to stop doing so or fix your code not to import both.

yjbdsky:
okey

PR #1201, opened 2023-12-13
Fix docker job

itizir:
FIx for the issue discussed here: https://github.com/mattn/go-sqlite3/pull/1177#issuecomment-1854438994

The second commit (changing the base image to be alpine 3.18, as done in https://github.com/authelia/authelia/pull/6404), wouldn't be needed once https://github.com/mattn/go-sqlite3/pull/1177 lands, I believe.


mattn:
Thank you

Issue #1215, opened 2024-02-05
Support for custom window functions


ohaibbq:
Hi, thanks for your development on this project!  I have a use case for registering a custom window function with the SQLite driver. 

Is there planned development for a `conn.RegisterWindowFunction` that uses `sqlite3_create_window_function` so new window functions can be written in Go?

https://www.sqlite.org/windowfunctions.html#udfwinfunc

PR #1216, opened 2024-02-07
WIP: Allow custom window functions to be registered

ohaibbq:
This is a draft PR to discuss the desired interface and implementation for #1215 

I've added a new function `RegisterWindowAggregator` that is similar to (read: copy/pasted from) `RegisterAggregator`. An alternative would be to check that the interface `impl` implements the two functions that windowing requires `Inverse` and `Value`.

I'm unsure if it makes sense to use the type-embedded `windowAggInfo`, or instead to consolidate this into `aggInfo`

Once I get clarification on these two points I can make this PR more review-ready

Thanks for taking a look!


ohaibbq:
closing in favor of a more "real" implementation!

PR #1220, opened 2024-02-11
feature: Allow custom window functions to be registered with the driver

### LINKED ISSUE #1215: Support for custom window functions

ohaibbq:
This implements the registration of user-defined window functions. It extends the existing `RegisterAggregator` interface to allow the `impl` to implement `Inverse` and `Value`, in which case, the aggregator can be used both as a regular aggregate function and a window function.

Closes #1215

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -42,14 +42,31 @@ func callbackTrampoline(ctx *C.sqlite3_context, argc int, argv **C.sqlite3_value
 //export stepTrampoline
 func stepTrampoline(ctx *C.sqlite3_context, argc C.int, argv **C.sqlite3_value) {
 	args := (*[(math.MaxInt32 - 1) / unsafe.Sizeof((*C.sqlite3_value)(nil))]*C.sqlite3_value)(unsafe.Pointer(argv))[:int(argc):int(argc)]
-	ai := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo)
-	ai.Step(ctx, args)
+	if ai, ok := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo); ok {
+		ai.Step(ctx, args)
+	}
+}
+
+//export inverseTrampoline
+func inverseTrampoline(ctx *C.sqlite3_context, argc C.int, argv **C.sqlite3_value) {
+	args := (*[(math.MaxInt32 - 1) / unsafe.Sizeof((*C.sqlite3_value)(nil))]*C.sqlite3_value)(unsafe.Pointer(argv))[:int(argc):int(argc)]
+	if ai, ok := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo); ok {
+		ai.Inverse(ctx, args)
+	}
+}
+
+//export valueTrampoline
+func valueTrampoline(ctx *C.sqlite3_context) {
+	if ai, ok := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo); ok {
+		ai.Value(ctx)
+	}
 }
 
 //export doneTrampoline
 func doneTrampoline(ctx *C.sqlite3_context) {
-	ai := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo)
-	ai.Done(ctx)
+	if ai, ok := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo); ok {
+		ai.Done(ctx)
+	}
 }
 
 //export compareTrampoline
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -159,8 +159,25 @@ int _sqlite3_create_function(
   return sqlite3_create_function(db, zFunctionName, nArg, eTextRep, (void*) pApp, xFunc, xStep, xFinal);
 }
 
+int _sqlite3_create_window_function(
+  sqlite3 *db,
+  const char *zFunctionName,
+  int nArg,
+  int eTextRep,
+  uintptr_t pApp,
+  void (*xStep)(sqlite3_context*,int,sqlite3_value**),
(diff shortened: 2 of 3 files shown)


ohaibbq:
Hi @mattn, let me know if there's anything further I can do to push this PR along. It is required for a project I am working on.

ohaibbq:
Hi @mattn, how can I be help in getting this merged?

mattn:
Sorry my delay. I'll look into it in later.

mattn:
Look good for me. @rittneje any opinion?

ohaibbq:
A friendly bump on this @rittneje

ohaibbq:
@mattn This would be very helpful to get in for the users of goccy/bigquery-emulator. I cannot merge my feature branch into the project as it uses a forked `go-sqlite3`.

Is there anything else I could do to push this along?

mattn:
Yes, I'm thinking this is useful but I want to wait @rittneje 's opinion.

ohaibbq:
Following up again!

ohaibbq:
@rittneje Did you have any thoughts on this small PR?

rittneje:
@ohaibbq Sorry, I really don't have the time for an in-depth review of this. (And I wouldn't describe this as a "small PR".)

sqlite3.go
```diff
@@ -719,93 +799,142 @@ func (c *SQLiteConn) RegisterAggregator(name string, impl any, pure bool) error
 	}
 
 	agg := t.Out(0)
+	var implReturnsPointer bool
 	switch agg.Kind() {
-	case reflect.Ptr, reflect.Interface:
+	case reflect.Ptr:
+		implReturnsPointer = true
+	case reflect.Interface:
+		implReturnsPointer = false
```
rittneje, line 807: 1. How can you conclude whether it returns a pointer or not in this case?
2. Saying it doesn't return a pointer and allowing that seems to contradict the error message in the default case.

sqlite3.go
```diff
@@ -719,93 +799,142 @@ func (c *SQLiteConn) RegisterAggregator(name string, impl any, pure bool) error
...
-		return errors.New("SQlite aggregator constructor must return a pointer object")
+		return errors.New("SQLite aggregator constructor must return a pointer object")
 	}
+
 	stepFn, found := agg.MethodByName("Step")
 	if !found {
-		return errors.New("SQlite aggregator doesn't have a Step() function")
+		return errors.New("SQLite aggregator doesn't have a Step() function")
+	}
+	err := ai.setupStepInterface(stepFn, &ai.stepArgConverters, &ai.stepVariadicConverter, implReturnsPointer, "Step()")
```
rittneje, line 816: The penultimate parameter to `setupStepInterface` is named `isImplPointer`. That doesn't seem to have anything to do with `implReturnsPointer`.

callback.go
```diff
@@ -42,14 +42,31 @@ func callbackTrampoline(ctx *C.sqlite3_context, argc int, argv **C.sqlite3_value
...
-	ai := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo)
-	ai.Step(ctx, args)
+	if ai, ok := lookupHandle(C.sqlite3_user_data(ctx)).(*aggInfo); ok {
+		ai.Step(ctx, args)
+	}
+}
+
+//export inverseTrampoline
+func inverseTrampoline(ctx *C.sqlite3_context, argc C.int, argv **C.sqlite3_value) {
+	args := (*[(math.MaxInt32 - 1) / unsafe.Sizeof((*C.sqlite3_value)(nil))]*C.sqlite3_value)(unsafe.Pointer(argv))[:int(argc):int(argc)]
```
rittneje, line 52: Can you explain what this is doing?

sqlite3.go
ohaibbq: You're right that I can't conclude if it returns a pointer or not in this case.

It does contradict the error message in the default case, but it was previously allowed in the initial implementation of user-defined functions #229.

Perhaps there is a better name for this variable.

sqlite3.go
ohaibbq: I could rename the parameter to `hasMethodReceiver`. 

I maintained the behavior from the original implementation, where we'd only skip the method receiver if the `agg.Kind() == reflect.Pointer`, but if we must always return a pointer, then there should always be a method receiver. 

According to the [`reflect.Type` docs](https://pkg.go.dev/reflect#Type):

>	// For a non-interface type T or *T, the returned Method's Type and Func
>	// fields describe a function whose first argument is the receiver,
>	// and only exported methods are accessible.


When running `sqlite3_test.go`, I did not encounter any cases where `implReturnsPointer` would be false. I wonder if we should remove this branch of the logic.

ohaibbq:
@rittneje I appreciate the time you've spent on your initial review of this. Perhaps I can clean up the confusing interface of allowing implementations that return `interface{}` when we only expect pointers. 

When testing registering an implementation that returns `interface{}`, registration understandably returns an error `SQLite aggregator doesn't have a Step() function`. 

If you don't have much time to continue review, maybe we can defer to @mattn's review since you've gotten the chance to gloss over the code.

ohaibbq:
@mattn any updates on this?

ohaibbq:
We'd love to get this in, it'd greatly improve the performance of https://github.com/goccy/bigquery-emulator.

ohaibbq:
We've migrated to the modernc.org driver

PR #1234, opened 2024-04-19
Add loongarch64 support.

abner-chenc:


mattn:
Thanks but sqlite3-binding.c is based on amalgamation code of sqlite3. So you have better to send your patch to sqlite3 developers.

PR #1249, opened 2024-06-13
Add loongarch64 byteorder

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.
Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html


charlievieth:
@zhaixiaojuan The `sqlite3-binding.c` file is not really part of this repo and is actually a copy of `sqlite3.c` from the sqlite3 [amalgamation](https://www.sqlite.org/download.html) distribution. 

TLDR: To make this change: it needs to happen in [sqlite3](https://www.sqlite.org/). Only then should the `sqlite3-binding.c` file be updated to reflect the latest release of the sqlite3.

PR #1254, opened 2024-07-01
Fix deliberate_fall_through macro for LLVM

r-barnes:
sqlite3-binding.c current has this code:
```
/*
** Macro to disable warnings about missing "break" at the end of a "case".
*/
#if GCC_VERSION>=7000000
# define deliberate_fall_through __attribute__((fallthrough));
#else
# define deliberate_fall_through
#endif
```
but this leaves LLVM without implicit fallthrough protections.

Can we use this instead?
```
#ifndef ATTRIBUTE_FALLTHROUGH
#  if defined __has_attribute
#    if __has_attribute (fallthrough)
#      define ATTRIBUTE_FALLTHROUGH __attribute__ ((fallthrough))
#    else
#      define ATTRIBUTE_FALLTHROUGH
#    endif
#  endif
#endif
```


rittneje:
We don't make changes directly to the C library. If and when your upstream patch makes it into an official release, it will be pulled in as part of that upgrade.

rittneje:
@r-barnes Once the change lands in an official SQLite release, we'll pull it in via the standard process. Consequently, I am going to decline this PR.

Please feel free to open an issue requesting an upgrade once the upstream release happens so we don't forget.

PR #1262, opened 2024-07-17
Fix sqlite3_opt_unlock_notify with USE_LIBSQLITE3

q66:
A valid sqlite header must always be included (like in the other files) but sqlite3-binding.h explicitly guards against the system library case.

This fixes forgejo compilation with libsqlite3 tag.


charlievieth:
@mattn / @rittneje this commit simply fixes the `sqlite3_opt_unlock_notify.{c,go}` files so that they respect `USE_LIBSQLITE3`. Now that Fedora is moving to Forgejo it might be nice to help them out.

charlievieth:
@rittneje thank you for merging this!!!

Issue #1266, opened 2024-08-06
Add Inline Documentation for adding Custom Extension Functions in `sqlite3-binding.c`


jeffreyaven:
## Background

We are using `mattn/go-sqlite3` in [`stackql/stackql`](https://github.com/stackql/stackql); we need to add custom extension functions specific to our program.  We wouldn't expect these to be committed to this project, but we want an easy, repeatable, automated method to add our functions in a fork via a `make` target (like `make amalgamate`).  We believe other developers would benefit from this capability as well.  

## Proposal

Inline documentation is proposed to facilitate developers' addition of custom extension functions.  My approach would be to add comments within the `sqlite3-binding.c` file that serve as placeholders for adding custom extension functions.  This would help external developers understand where to include their extension functions without requiring the maintainers of `matte/go-sqlite3` to verify, test, or maintain those external functions.

### 1.  Forward Declarations

```c
// CUSTOM_EXTENSIONS
// 1.  Add custom extension initializer function declarations here...
// End CUSTOM_EXTENSIONS
```

### 2.  Extension Initializer Array

```c
// CUSTOM_EXTENSIONS
// 2.  Include custom extension initializer functions here...
// End CUSTOM_EXTENSIONS
```

### 3.  Extension Initializer Functions

```c
// CUSTOM_EXTENSIONS
// 3.  Add custom extension initializer functions here...
// End CUSTOM_EXTENSIONS
```

## Benefit

These comments will provide clear guidance to developers looking to add custom extensions, helping them to understand the appropriate sections in the `sqlite3-binding.c` file where they should include their code.

## Request

I would like to open a pull request to add these comments and would appreciate any feedback or suggestions.

PR #1267, opened 2024-08-06
Add inline documentation for custom extension functions

jeffreyaven:
closes https://github.com/mattn/go-sqlite3/issues/1266


rittneje:
This is the SQLite amalgamation file. We do not make edits to it. If you want this change made, you will need to get the SQLite authors to add it to the upstream.

PR #1278, opened 2024-10-01
Add a new `RegisterBusyHandler` method to `SQLiteConn` to allow access to `sqlite3_busy_handler()`

polyscone:
This PR just adds a new `RegisterBusyHandler` method to the `SQLiteConn` type to allow use of the `sqlite3_busy_handler()` function.

I tried to follow the same style as the other `Register*` methods which use the `void*` user data args as a way to pass a handle to a Go callback function to a trampoline function defined in C.

I've also updated the `_examples` folder with a new example file.


polyscone:
@mattn Sorry if you're busy, but is there any chance I could bother you to take a look at this, please?

polyscone:
Is there any chance I could get someone to take a look at this? @mattn , or maybe @rittneje ?

PR #1285, opened 2024-10-11
Return io.EOF for empty Query statements

ganigeorgiev:
This small PR is supposed to fix https://github.com/mattn/go-sqlite3/issues/950.


rittneje:
I don't think we should return `io.EOF`. An empty query is likely a bug, and doing this would mask it.

ganigeorgiev:
It is an empty statement. I don't see why it would be considered bug.

I have no strong opinion on this but note that https://pkg.go.dev/modernc.org/sqlite also seems to behave this way and doesn't hang.

ganigeorgiev:
This also seems close to the behavior of the already merged https://github.com/mattn/go-sqlite3/pull/973.

PR #1287, opened 2024-10-22
question: why do the if checks do the same thing.

svanellewee:
This is more a question than an MR. Why does the macro `#if` checks do the same thing in both branches ?

Would it not make sense to have each branch do something different ?

Sorry was trying to build gogs on alpine and noticed this. (This seems to work on alpine if I try it though)


rittneje:
This is the actual C code from the SQLite authors, which we don’t tamper with. You should ask on the forum if you want to know why they did it that way.

svanellewee:
Thank you for the response. Is there a way of building this project on platforms like alpine that avoids this issue ?

rittneje:
Which version of this library are you using?

PR #1295, opened 2024-11-07
Improve the performance of QueryContext by reusing the result channel

charlievieth:
This commit improves the performance of QueryContext by changing it to reuse the result channel instead of creating a new one for each query. This is particularly impactful for queries that scan more than one row.

It also adds a test that actually exercises the sqlite3_interrupt logic since the existing tests did not. Those tests cancelled the context before scanning any of the rows and could be made to pass without ever calling sqlite3_interrupt. The below version of SQLiteRows.Next passes the previous tests:

```go
func (rc *SQLiteRows) Next(dest []driver.Value) error {
	rc.s.mu.Lock()
	defer rc.s.mu.Unlock()
	if rc.s.closed {
		return io.EOF
	}
	if err := rc.ctx.Err(); err != nil {
		return err
	}
	return rc.nextSyncLocked(dest)
}
```

Benchmark results:
```
goos: darwin
goarch: arm64
cpu: Apple M1 Max
                                          │    b.txt    │               n.txt                │
                                          │   sec/op    │   sec/op     vs base               │
Suite/BenchmarkQueryContext/Background-10   4.088µ ± 1%   4.154µ ± 3%  +1.60% (p=0.011 n=10)
Suite/BenchmarkQueryContext/WithCancel-10   12.84µ ± 3%   11.67µ ± 3%  -9.08% (p=0.000 n=10)
geomean                                     7.245µ        6.963µ       -3.89%

                                          │    b.txt     │                 n.txt                  │
                                          │     B/op     │     B/op      vs base                  │
Suite/BenchmarkQueryContext/Background-10     400.0 ± 0%     400.0 ± 0%        ~ (p=1.000 n=10) ¹
Suite/BenchmarkQueryContext/WithCancel-10   2.547Ki ± 0%   1.282Ki ± 0%  -49.67% (p=0.000 n=10)
geomean                                      1021.4          724.6       -29.06%
¹ all samples are equal

                                          │   b.txt    │                n.txt                 │
                                          │ allocs/op  │ allocs/op   vs base                  │
Suite/BenchmarkQueryContext/Background-10   12.00 ± 0%   12.00 ± 0%        ~ (p=1.000 n=10) ¹
Suite/BenchmarkQueryContext/WithCancel-10   49.00 ± 0%   28.00 ± 0%  -42.86% (p=0.000 n=10)
geomean                                     24.25        18.33       -24.41%
¹ all samples are equal
```


charlievieth:
@mattn / @rittneje would either of you be able to take a look at this PR?

PR #1296, opened 2024-11-07
Fix exponential memory allocation in SQLiteConn.Exec

charlievieth:
This commit fixes SQLiteConn.Exec to not duplicate the provided SQL statement and to pass the Go string to libsqlite3 instead of creating a C string copy and passing that. The issue with the previous implementation was that it would create a new C string copy of the original Go string each time it stepped through a SQL statement. This led to exponential memory usage when Exec'ing an argument that contained multiple SQL statements (which is common when initializing a database from a SQL dump).

This commit is a slimmed down version of https://github.com/mattn/go-sqlite3/pull/1133

```
goos: darwin
goarch: arm64
cpu: Apple M1 Max
                           │    b.txt     │                n.txt                │
Suite/BenchmarkExec-10        1.278µ ± 1%   1.022µ ± 2%  -19.99% (p=0.000 n=10)
Suite/BenchmarkExecStep-10   1762.9µ ± 0%   900.9µ ± 1%  -48.90% (p=0.000 n=10)
geomean                       47.47µ        30.35µ       -36.06%

                           │     b.txt      │                n.txt                 │
                           │      B/op      │     B/op      vs base                │
Suite/BenchmarkExec-10           128.0 ± 0%     120.0 ± 0%   -6.25% (p=0.000 n=10)
Suite/BenchmarkExecStep-10   5279.94Ki ± 0%   31.34Ki ± 0%  -99.41% (p=0.000 n=10)
geomean                        25.69Ki        1.916Ki       -92.54%

                           │    b.txt    │                n.txt                │
                           │  allocs/op  │  allocs/op   vs base                │
Suite/BenchmarkExec-10        7.000 ± 0%    6.000 ± 0%  -14.29% (p=0.000 n=10)
Suite/BenchmarkExecStep-10   7.000k ± 0%   3.003k ± 0%  -57.10% (p=0.000 n=10)
geomean                       221.4         134.2       -39.36%
```


charlievieth:
**Note:** This PR has been updated to include logic for Exec'ing queries that have no bind parameters.

charlievieth:
@rittneje Any chance I could get this reviewed?

mattn:
@rittneje Could you please code-review this?

unsafe_go121.go
```diff
@@ -0,0 +1,23 @@
+//go:build go1.21
+// +build go1.21
+
+// The unsafe.StringData function was made available in Go 1.20 but it
+// was not until Go 1.21 that Go was changed to interpret the Go version
+// in go.mod (1.19 as of writing this) as the minimum version required
```
rittneje, line 6: @mattn Currently, the GitHub workflow tests 1.19, 1.20, and 1.21, so it is pretty out of date. And the readme links to the offical Golang release policy, which (at this moment) only supports 1.23 and 1.24. Should we just change go.mod and GitHub workflow to 1.23+?

unsafe_go121.go
```diff
@@ -0,0 +1,23 @@
+//go:build go1.21
+// +build go1.21
+
+// The unsafe.StringData function was made available in Go 1.20 but it
```
rittneje, line 4: I'm confused - if it was added in 1.20, why isn't the build constraint `go1.20`? The go mod directive historically was about enabling language features, not library functions.

unsafe_go121.go
```diff
@@ -0,0 +1,23 @@
...
+import "unsafe"
+
+// stringData is a safe version of unsafe.StringData that handles empty strings.
+func stringData(s string) *byte {
+	if len(s) != 0 {
+		return unsafe.StringData(s)
+	}
+	// The return value of unsafe.StringData
+	// is unspecified if the string is empty.
+	return &placeHolder[0]
```
rittneje, line 22: I don't quite follow why this is necessary. If the length of the string is empty, why does it matter (to C) what pointer we give it?

sqlite3.go
```diff
@@ -137,6 +137,61 @@ _sqlite3_prepare_v2_internal(sqlite3 *db, const char *zSql, int nBytes, sqlite3_
...
+			return rv;
+		}
+
+		// Process statement
+		do {
+			rv = _sqlite3_step_internal(stmt);
+		} while (rv == SQLITE_ROW);
+
+		// Only record the number of changes made by the last statement.
+		*changes = sqlite3_changes64(db);
```
rittneje, line 177: Should we really do this if there was an error?

sqlite3.go
```diff
@@ -137,6 +137,61 @@ _sqlite3_prepare_v2_internal(sqlite3 *db, const char *zSql, int nBytes, sqlite3_
...
+		*rowid = sqlite3_last_insert_rowid(db);
+
+		if (rv != SQLITE_OK && rv != SQLITE_DONE) {
+			sqlite3_finalize(stmt);
+			return rv;
+		}
+		rv = sqlite3_finalize(stmt);
```
rittneje, line 184: > If the most recent evaluation of the statement encountered no errors or if the statement is never been evaluated, then sqlite3_finalize() returns SQLITE_OK. If the most recent evaluation of statement S failed, then sqlite3_finalize(S) returns the appropriate [error code](https://www.sqlite.org/rescode.html) or [extended error code](https://www.sqlite.org/rescode.html#extrc).

This implies that there is no need to inspect the return value here, and it can be done unconditionally (before checking `rv`).

sqlite3.go
```diff
@@ -137,6 +137,61 @@ _sqlite3_prepare_v2_internal(sqlite3 *db, const char *zSql, int nBytes, sqlite3_
...
+		if (rv != SQLITE_OK) {
+			return rv;
+		}
+
+		nBytes -= tail - zSql;
```
rittneje, line 189: Here you are assuming `tail != NULL` but above in `_sqlite3_prepare_v2` you have explicit handling for that. I think we should be consistent.

sqlite3.go
```diff
@@ -858,54 +913,119 @@ func (c *SQLiteConn) Exec(query string, args []driver.Value) (driver.Result, err
 }
 
 func (c *SQLiteConn) exec(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
-	start := 0
+	// Trim the query. This is mostly important for getting rid
```
rittneje, line 916: Why?

sqlite3.go
```diff
@@ -858,54 +913,119 @@ func (c *SQLiteConn) Exec(query string, args []driver.Value) (driver.Result, err
...
+					stmtArgs = append(stmtArgs, args[i])
 				}
 			}
-			res, err = s.(*SQLiteStmt).exec(ctx, stmtArgs)
+			for i := range stmtArgs {
+				stmtArgs[i].Ordinal = i + 1
+			}
+			var err error
+			res, err = s.exec(ctx, stmtArgs)
 			if err != nil && err != driver.ErrSkip {
```
rittneje, line 963: What does it mean if `exec` returns `ErrSkip`, and why is it safe to ignore it here?

sqlite3.go
```diff
@@ -858,54 +913,119 @@ func (c *SQLiteConn) Exec(query string, args []driver.Value) (driver.Result, err
...
-		if err != nil {
-			return nil, err
+		s = SQLiteStmt{c: c} // reset
+		sz = 0
+		rv := C._sqlite3_prepare_v2(c.db, (*C.char)(unsafe.Pointer(stringData(query))),
+			C.int(len(query)), &s.s, &sz)
+		if rv != C.SQLITE_OK {
+			return nil, c.lastError()
 		}
+		query = strings.TrimSpace(query[sz:])
```
rittneje, line 940: Why `TrimSpace`?

sqlite3.go
```diff
@@ -858,54 +913,119 @@ func (c *SQLiteConn) Exec(query string, args []driver.Value) (driver.Result, err
...
+				s.finalize()
 				return nil, err
 			}
 			start += na
 		}
-		tail := s.(*SQLiteStmt).t
-		s.Close()
-		if tail == "" {
+		s.finalize()
+		if len(query) == 0 {
```
rittneje, line 970: I think there is an important nuance in play here. There are two ways to know you are done. The first is that `pzTail` is the empty string. The second is that `ppStmt` is null.

However, the flaw in the current code is that it assumes that just checking whether `pzTail` is whitespace is a sufficient stand-in. This is not the case, because you could have a trailing comment like `"DELETE FROM MyTable; -- foo"`. In this case, `pzTail` will be `" -- foo"` after the first iteration, and then `ppStmt` will be null after the second iteration. But as written, I don't think that case is handled correctly by this code.

(Please also add a test case for the trailing comment situation.)

sqlite3_test.go
```diff
@@ -1090,6 +1091,67 @@ func TestExecer(t *testing.T) {
 	}
 }
 
+func TestExecDriverResult(t *testing.T) {
+	setup := func(t *testing.T) *sql.DB {
```
rittneje, line 1095: This is only called once - can we just merge it into `test`?

rittneje (changes requested):
see other comments

PR #1298, opened 2024-11-22
Add ability to set an int64 file control

jonstacks:
Similar to https://github.com/mattn/go-sqlite3/pull/1000,

We are using `vfs=memdb` and would like to use the `SQLITE_FCNTL_SIZE_LIMIT` opcode with `sqlite3_file_control`:

> The [SQLITE_FCNTL_SIZE_LIMIT](https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntlsizelimit) opcode is used by in-memory VFS that implements [sqlite3_deserialize()](https://www.sqlite.org/c3ref/deserialize.html) to set an upper bound on the size of the in-memory database. The argument is a pointer to a [sqlite3_int64](https://www.sqlite.org/c3ref/int64.html). If the integer pointed to is negative, then it is filled in with the current limit. Otherwise the limit is set to the larger of the value of the integer pointed to and the current database size. The integer pointed to is set to the new limit.

Calling the existing `SetFileControlInt` method when go decides that `int`s are `int64` based on your architecture seem to overflow the `C.int` and in our case set the size limit of our in-memory VFS to 0. This adds a new `SetFileControlInt64` to be used with op codes that expect a `sqlite3_int64`.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1876,6 +1876,9 @@ func (c *SQLiteConn) SetLimit(id int, newVal int) int {
 // This method is not thread-safe as the returned error code can be changed by
 // another call if invoked concurrently.
 //
+// Use SetFileControlInt64 instead if the argument for the opcode is documented
+// as a pointer to a sqlite3_int64.
+//
 // See: sqlite3_file_control, https://www.sqlite.org/c3ref/file_control.html
 func (c *SQLiteConn) SetFileControlInt(dbName string, op int, arg int) error {
 	if dbName == "" {
@@ -1901,8 +1904,8 @@ func (c *SQLiteConn) SetFileControlInt(dbName string, op int, arg int) error {
 // This method is not thread-safe as the returned error code can be changed by
 // another call if invoked concurrently.
 //
-// Prefer this method over SetFileControlInt when the argument to the underlying
-// SQLite function is an int64.
+// Only use this method if the argument for the opcode is documented as a pointer
+// to a sqlite3_int64.
 //
 // See: sqlite3_file_control, https://www.sqlite.org/c3ref/file_control.html
 func (c *SQLiteConn) SetFileControlInt64(dbName string, op int, arg int64) error {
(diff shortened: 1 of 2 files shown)


sqlite3.go
```diff
@@ -1893,6 +1893,34 @@ func (c *SQLiteConn) SetFileControlInt(dbName string, op int, arg int) error {
...
 
+// SetFileControlInt invokes the xFileControl method on a given database. The
+// dbName is the name of the database. It will default to "main" if left blank.
+// The op is one of the opcodes prefixed by "SQLITE_FCNTL_". The arg argument
+// and return code are both opcode-specific. Please see the SQLite documentation.
+//
+// This method is not thread-safe as the returned error code can be changed by
+// another call if invoked concurrently.
+//
+// Prefer this method over SetFileControlInt when the argument to the underlying
```
rittneje, line 1904: I think this method should say "only use if the argument is a pointer to a sqlite3_int64". Meanwhile `SetFileControlInt` should say "useSetFileControlInt64 instead if the argument is a pointer to a sqlite3_int64".

rittneje:
> Calling the existing SetFileControlInt method when go decides that ints are int64 based on your architecture seem to overflow the C.int and in our case set the size limit of our in-memory VFS to 0.

I filed #1299 for this bug.

sqlite3.go
jonstacks: Good call out. I've updated the documentation in https://github.com/mattn/go-sqlite3/pull/1298/commits/9882f8f441bfda9333ad5f70119d5692ff960694

sqlite3.go
```diff
@@ -1893,6 +1896,34 @@ func (c *SQLiteConn) SetFileControlInt(dbName string, op int, arg int) error {
 	return nil
 }
 
+// SetFileControlInt invokes the xFileControl method on a given database. The
```
rittneje, line 1899:
```suggestion
// SetFileControlInt64 invokes the xFileControl method on a given database. The
```

jonstacks:
CIFuzz failure was caused by https://github.com/mattn/go-sqlite3/pull/1300. I'll rebase this and hopefully that fixes the issue 🤞

rittneje (approved):
LGTM
@mattn any concerns on your end?

PR #1300, opened 2024-11-26
Upgrade upload-artifact action

jonstacks:
On #1298, the [fuzzing action failed to run](https://github.com/mattn/go-sqlite3/actions/runs/12014833442/job/33492262820) 

image

action/upload-artifact@v1 has been deprecated for a while. It seems like GitHub Actions will now cancel workflows if it is still using v1 of the action. This upgrades to the latest v4 of the action.

See https://github.blog/news-insights/product-news/get-started-with-v4-of-github-actions-artifacts/ for details.


mattn:
Thank you

PR #1302, opened 2024-12-09
always close db connection on error in Open

charlievieth:
This commit fixes a bug where the db connection would not always be closed when an error occurred in Open. Additionally, it makes sure that we always unregister any callbacks associated with the connection, which previously did not always happen. This change consolidates the error handling logic (previously, it had to be done for each return statement) which should make it more difficult to introduce this type of bug in the future.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1487,296 +1487,294 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 		return nil
 	}
 
-	// Busy timeout
-	if err := exec(fmt.Sprintf("PRAGMA busy_timeout = %d;", busyTimeout)); err != nil {
-		C.sqlite3_close_v2(db)
-		return nil, err
-	}
-
-	// USER AUTHENTICATION
-	//
-	// User Authentication is always performed even when
-	// sqlite_userauth is not compiled in, because without user authentication
-	// the authentication is a no-op.
-	//
-	// Workflow
-	//	- Authenticate
-	//		ON::SUCCESS		=> Continue
-	//		ON::SQLITE_AUTH => Return error and exit Open(...)
-	//
-	//  - Activate User Authentication
-	//		Check if the user wants to activate User Authentication.
-	//		If so then first create a temporary AuthConn to the database
-	//		This is possible because we are already successfully authenticated.
-	//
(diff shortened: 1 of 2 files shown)


mattn:
Please fix conflict

charlievieth:
Conflict fixed and thank you for the quick review!

sqlite3.go
```diff
@@ -1487,296 +1487,294 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
...
-			if err := conn.RegisterFunc("sqlite_crypt", CryptEncoderSSHA512(authSalt), true); err != nil {
-				return nil, fmt.Errorf("CryptEncoderSSHA512: %s", err)
-			}
+	// All further database configuration *must* occur within this function, as
+	// it simplifies closing the database connection and unregistering associated
+	// functions or callbacks.
+	//
+	// Previously, this was handled at each return statement, which was error-prone
+	// and led to bugs because the database connection was not always closed.
+	err := func() error {
```
rittneje, line 1499: Instead of this, we can use a named return value and a defer. That will also enable closing it on panic, which is another hole we have today.
```go
defer func() {
    if retErr != nil {
        conn.Close()
    }
}()
```

rittneje: Sorry, I was a little unclear. I just meant that using a `defer` would enable accounting for panics, not the named return itself.

The reason to close in the panic case is that this library cannot control what the caller of `Open` does. In particular, they could choose to swallow panics, and it would be good to at least avoid causing a memory leak in that situation.

charlievieth: > Sorry, I was a little unclear. I just meant that using a defer would enable accounting for panics, not the named return itself.

Thanks and yeah was a little confused since in the given code snipped the defer would not close the db conn in the event of a panic (since the named return value would likely be nil). 

I updated the code to close the db connection if a panic occurs during setup.

charlievieth: @rittneje I updated the code to incorporate your comments any chance you could take another look at this?

charlievieth: @rittneje Is there anything that I can do to help get this over the line and merged?

rittneje (approved):
LGTM

PR #1320, opened 2025-02-25
Add sqlite_dbpage extension build tag

hifi:
Provides raw access to database pages. Useful for taking consistent live snapshots of WAL databases without causing write locks by reading out all of the pages and writing them either to a file or some external storage.

The [sqlite3_rsync](https://www.sqlite.org/rsync.html) utility that is bundled with sqlite3 source is one way to use it to somewhat efficiently synchronize a remote database.


hifi:
Added a test that makes a full copy of a database using sqlite_dbpage table and checks the copy is consistent and has the expected data.

Issue #600, opened 2018-06-25
How do I get the type of the function field in the return result?


jasonGeng88:
in mysql case:
![image](https://user-images.githubusercontent.com/11816184/41843411-0ef8e1dc-78a0-11e8-9a47-a347925032b4.png)

![image](https://user-images.githubusercontent.com/11816184/41843190-78c1ce40-789f-11e8-80fb-fc4ff754c039.png)

in sqlite3 case:
<img width="579" alt="2018-06-25 5 48 18" src="https://user-images.githubusercontent.com/11816184/41844003-83fc461c-78a1-11e8-89a4-9c2eb599e382.png">

![image](https://user-images.githubusercontent.com/11816184/41843484-359b7160-78a0-11e8-9bd9-34f21944a572.png)

I saw the #395 issue. Is this the same problem?

GJRTimmer:
Actually I'm running into this while trying to write a TestCase for the `ColumnTypeScanType` function.

GJRTimmer:
You need to request the ColumnTypes from within a Range over `rows`. This is due to how SQLite works. See [sqlite.org](https://www.sqlite.org/c3ref/column_blob.html) for more information. SQLite has to have `_step()` called before it can return what you looking for.

If anything is unclear take a peek at the v2.0.0 within result_test.go you can find what you are looking for.

PR #1327, opened 2025-03-17
use sqlite3_column_type() for ColumnTypeScanType()

alkemir:
The current implementation was changed in (PR-909)[https://github.com/mattn/go-sqlite3/pull/909] stating that  `sqlite3_column_type()` always returns SQLITE_NULL, but as noted in (ISSUE-600)[https://github.com/mattn/go-sqlite3/issues/600], the case was probably because for SQLite3, `sqlite3_column_type()` must be called after calling Next(). sqlite3_column_type() returns the types for the columns on a given row, thus the cursor must be pointing to a valid row. See SQLite docs on API (lifecycle)[https://www.sqlite.org/cintro.html], (rows and columns methods)[https://www.sqlite.org/c3ref/column_blob.html] and (this forum post)[https://sqlite.org/forum/forumpost/2e582c74b6215555?t=c&unf]

The [Go API](https://pkg.go.dev/database/sql/driver#RowsColumnTypeScanType) does not restrain the behavior of `RowsColumnTypeScanType()` as to when it should be called. Thus, the implementation in this PR makes a compromise and falls back to `sqlite3_column_decltype()` in a best effort to remain retro-compatible, but this is still a potentially breaking change.

The advantage of using `sqlite3_column_type()` over `sqlite3_column_decltype()` is two-fold:

1. It is closer to the SQLite3 API 
1. It will return the correct scanType for expressions and aggregate functions, which are the cases in which `sqlite3_column_decltype()` is insufficient. Regarding this aspect, the existing implementation provides no way to access the values without potentially triggering a type conversion, unless the application is aware of the result type of the query beforehand.

### DIFF
--- a/sqlite3_type.go
+++ b/sqlite3_type.go
@@ -18,6 +18,27 @@ import (
 	"strings"
 )
 
+const (
+	SQLITE_INTEGER = iota
+	SQLITE_TEXT
+	SQLITE_BLOB
+	SQLITE_REAL
+	SQLITE_NUMERIC
+	SQLITE_TIME
+	SQLITE_BOOL
+	SQLITE_NULL
+)
+
+var (
+	TYPE_NULLINT    = reflect.TypeOf(sql.NullInt64{})
+	TYPE_NULLFLOAT  = reflect.TypeOf(sql.NullFloat64{})
+	TYPE_NULLSTRING = reflect.TypeOf(sql.NullString{})
+	TYPE_RAWBYTES   = reflect.TypeOf(sql.RawBytes{})
+	TYPE_NULLBOOL   = reflect.TypeOf(sql.NullBool{})
+	TYPE_NULLTIME   = reflect.TypeOf(sql.NullTime{})
+	TYPE_ANY        = reflect.TypeOf(new(any))
+)
+
 // ColumnTypeDatabaseTypeName implement RowsColumnTypeDatabaseTypeName.
 func (rc *SQLiteRows) ColumnTypeDatabaseTypeName(i int) string {
 	return C.GoString(C.sqlite3_column_decltype(rc.s.s, C.int(i)))
@@ -39,42 +60,47 @@ func (rc *SQLiteRows) ColumnTypeNullable(i int) (nullable, ok bool) {
 }
 
 // ColumnTypeScanType implement RowsColumnTypeScanType.
+// In SQLite3, this method should be called after Next() has been called, as sqlite3_column_type()
+// returns the column type for a specific row. If Next() has not been called, fallback to
+// sqlite3_column_decltype()
 func (rc *SQLiteRows) ColumnTypeScanType(i int) reflect.Type {
-	//ct := C.sqlite3_column_type(rc.s.s, C.int(i))  // Always returns 5
+	switch C.sqlite3_column_type(rc.s.s, C.int(i)) {
+	case C.SQLITE_INTEGER:
+		return TYPE_NULLINT
+	case C.SQLITE_FLOAT:
+		return TYPE_NULLFLOAT
+	case C.SQLITE_TEXT:
+		return TYPE_NULLSTRING
+	case C.SQLITE_BLOB:
+		return TYPE_RAWBYTES
+		//case C.SQLITE_NULL:
+		//	return TYPE_ANY
+	}
+
+	// Fallback to schema declared to remain retro-compatible
 	return scanType(C.GoString(C.sqlite3_column_decltype(rc.s.s, C.int(i))))
 }
-
-const (
-	SQLITE_INTEGER = iota
-	SQLITE_TEXT
-	SQLITE_BLOB
-	SQLITE_REAL
-	SQLITE_NUMERIC
-	SQLITE_TIME
-	SQLITE_BOOL
-	SQLITE_NULL
-)
 
 func scanType(cdt string) reflect.Type {
 	t := strings.ToUpper(cdt)
 	i := databaseTypeConvSqlite(t)
 	switch i {
 	case SQLITE_INTEGER:
-		return reflect.TypeOf(sql.NullInt64{})
+		return TYPE_NULLINT
 	case SQLITE_TEXT:
-		return reflect.TypeOf(sql.NullString{})
+		return TYPE_NULLSTRING
 	case SQLITE_BLOB:
-		return reflect.TypeOf(sql.RawBytes{})
+		return TYPE_RAWBYTES
 	case SQLITE_REAL:
-		return reflect.TypeOf(sql.NullFloat64{})
+		return TYPE_NULLFLOAT
 	case SQLITE_NUMERIC:
-		return reflect.TypeOf(sql.NullFloat64{})
+		return TYPE_NULLFLOAT
 	case SQLITE_BOOL:
-		return reflect.TypeOf(sql.NullBool{})
+		return TYPE_NULLBOOL
 	case SQLITE_TIME:
-		return reflect.TypeOf(sql.NullTime{})
+		return TYPE_NULLTIME
 	}
-	return reflect.TypeOf(new(any))
+	return TYPE_ANY
 }
 
 func databaseTypeConvSqlite(t string) int {


sqlite3_type.go
```diff
@@ -39,42 +60,47 @@ func (rc *SQLiteRows) ColumnTypeNullable(i int) (nullable, ok bool) {
...
+	switch C.sqlite3_column_type(rc.s.s, C.int(i)) {
+	case C.SQLITE_INTEGER:
+		return TYPE_NULLINT
+	case C.SQLITE_FLOAT:
+		return TYPE_NULLFLOAT
+	case C.SQLITE_TEXT:
+		return TYPE_NULLSTRING
+	case C.SQLITE_BLOB:
+		return TYPE_RAWBYTES
+		//case C.SQLITE_NULL:
```
otoolep, line 76: Why is this commented out?

alkemir: Its to document that the omission is intentional and not a miss, this is the actual value that we would be getting in 2 cases:

1. If we call before we have a valid row
1. If the value for the row-column is NULL

Happy to remove it or add a comment clarifying.

otoolep: It looks like a mistake. An explicit comment is better.

alkemir: Added comment. Please take another look.

otoolep: 👍

sqlite3_type.go
```diff
@@ -18,6 +18,27 @@ import (
...
+	SQLITE_TEXT
+	SQLITE_BLOB
+	SQLITE_REAL
+	SQLITE_NUMERIC
+	SQLITE_TIME
+	SQLITE_BOOL
+	SQLITE_NULL
+)
+
+var (
```
rittneje, line 32: Why are these all public?

sqlite3_type.go
```diff
@@ -39,42 +60,50 @@ func (rc *SQLiteRows) ColumnTypeNullable(i int) (nullable, ok bool) {
 }
 
 // ColumnTypeScanType implement RowsColumnTypeScanType.
+// In SQLite3, this method should be called after Next() has been called, as sqlite3_column_type()
+// returns the column type for a specific row. If Next() has not been called, fallback to
+// sqlite3_column_decltype()
 func (rc *SQLiteRows) ColumnTypeScanType(i int) reflect.Type {
-	//ct := C.sqlite3_column_type(rc.s.s, C.int(i))  // Always returns 5
+	switch C.sqlite3_column_type(rc.s.s, C.int(i)) {
```
rittneje, line 67: > If the SQL statement does not currently point to a valid row, or if the column index is out of range, the result is undefined.

We cannot call this if we haven't called `Next` yet, or if we have iterated past the last row, even if it happens to work today.

rittneje:
Please add a unit test for the new behavior.

sqlite3_type.go
alkemir: I am not sure if you mean 
1. `SQLiteRows.ColumnTypeScanType()` should make sure that we have a valid row so it does not run into undefined behavior, in which case it should signal an error condition, or
1. The driver should not call `C.sqlite3_column_type()` except inside `Next()`, which is the only point at which we can assure that we have a valid row.

If it is the first, the method signature does not allow a clear way of signaling errors (except for the empty interface, which signals not-supported) and I can't find a way to know if we have a valid row. The design of the packages (sql and go-sqlite3) seems to imply this should not be a concern beyond `Next()` and `Scan()`. Also the docs do not prescribe a context in which it is ok to call this method, but they do for other methods like `Close()` and `Scan()` implying it should always be ok to call `SQLiteRows.ColumnTypeScanType()`. The risk here is that the undefined behavior might mask as a type that is actually invalid for scanning. But, the implication is that the calling code would then try to `Scan()`, at which point it would get an error, so I see the risk as mitigated.

If it is the second, it is troublesome because that means there is no way of getting the rows without risking a type conversion unless the types are known by the calling code.

sqlite3_type.go
```diff
@@ -18,6 +18,27 @@ import (
...
+)
+
+var (
+	TYPE_NULLINT    = reflect.TypeOf(sql.NullInt64{})
+	TYPE_NULLFLOAT  = reflect.TypeOf(sql.NullFloat64{})
+	TYPE_NULLSTRING = reflect.TypeOf(sql.NullString{})
+	TYPE_RAWBYTES   = reflect.TypeOf(sql.RawBytes{})
+	TYPE_NULLBOOL   = reflect.TypeOf(sql.NullBool{})
+	TYPE_NULLTIME   = reflect.TypeOf(sql.NullTime{})
+	TYPE_ANY        = reflect.TypeOf(new(any))
```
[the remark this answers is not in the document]
alkemir: `*any` is the expected type, see the `Rows.Scan()` [docs](https://pkg.go.dev/database/sql#Rows.Scan):

> // If an argument has type *interface{}, Scan copies the value
> // provided by the underlying driver without conversion. When scanning
> // from a source value of type []byte to *interface{}, a copy of the
> // slice is made and the caller owns the result.

Relevant code:
```
case nil:
    switch d := dest.(type) {
    case *any:
	if d == nil {
		return errNilPtr
	}
	*d = nil
	return nil
```

In our case, this works:

```
	for rr.Next() {
		cc, err := rr.ColumnTypes()
		if err != nil {
			log.Fatal(err)
		}

		pointers := make([]any, 0)
		for _, ct := range cc {
			fmt.Printf("%s ", ct.ScanType())
			pointers = append(pointers, reflect.New(ct.ScanType()).Interface())
		}
		fmt.Println()

		if err := rr.Scan(pointers...); err != nil {
			log.Fatal(err)
		}

		fmt.Println(pointers...)
		for _, p := range pointers {
			if pn, ok := p.(**any); ok {
				fmt.Printf("%t ", *pn == nil)
			}
		}

	}
```

>sql.NullInt64 sql.NullString sql.NullFloat64 
>&{1 true} &{hello true} &{3.1415 true}
>sql.NullInt64 *interface {} *interface {}
>&{2 true} 0xc000064090 0xc000064098
>true true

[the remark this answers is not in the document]
rittneje: That's unrelated. `Rows.Scan` is talking about its arguments, which by necessity all have to be pointers as they are essentially output parameter. Meanwhile, `ColumnTypeScanType` is talking about the type to scan into (so not a pointer).

For example, `ColumnTypeScanType` mentions returning `reflect.TypeOf(int64(0))` while `Rows.Scan` mentions `*int64`.

You'll also notice that all the other types being returned (e.g., `TYPE_NULLSTRING`) here are not pointers.

In other words, you are supposed to create a variable of the type from `ColumnTypeScanType`, and then pass a pointer to it to `Rows.Scan`.

sqlite3_type.go
rittneje: This is a method on `SQLiteRows`, which certainly has the ability to know whether `Next` has been called yet, and whether the last call returned `io.EOF`. All it needs is some bookkeeping.

Also, "undefined behavior" means that anything could happen, including an application crash, now or in the future.

sqlite3_type.go
alkemir: Just had my Java brain on. Making them private.

sqlite3_type.go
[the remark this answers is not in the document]
alkemir: You are right. I will return `interface{}` now, following [Go 1.21](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/database/sql/sql.go;l=3214) , although this has been changed for [Go 1.24](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/database/sql/sql.go) for compatibility.

Still, this might break code depending on the previous behavior.

alkemir:
Should we still return the nullable types? When SQLite returns a NULL value, the column type will be type_any, so there is no need for nullable types now.

sqlite3_type.go
alkemir: Added guards. Thank you for pointing out `sqlite3_stmt_busy`

alkemir:
Thank you for all the reviews, is there anything else that I can do to improve this PR?

mattn:
Please add unit tests

alkemir:
Unit tests are already included, is the PR missing any particular aspect?

mattn:
Thank you for this PR. Using `sqlite3_column_type()` is the right direction — the current `sqlite3_column_decltype()` approach cannot handle expressions and aggregate functions. However, I have some concerns I'd like to discuss before merging:

- **Return type change from nullable to non-nullable**: `sql.NullInt64` → `int64`, `sql.NullString` → `string`, etc. Some downstream code that uses `reflect.New(ct.ScanType())` and type-asserts on the result may break.
- **Exported constants removed**: `SQLITE_INTEGER`, `SQLITE_TEXT`, `SQLITE_BLOB`, `SQLITE_REAL`, `SQLITE_NUMERIC`, `SQLITE_TIME`, `SQLITE_BOOL`, `SQLITE_NULL` are deleted. Any downstream code referencing these will fail to compile.
- **Behavior now depends on call timing**: Before `Next()` it returns `interface{}`, after `Next()` it returns the actual type. The old implementation always returned based on the declared type regardless of cursor position.
- **Loss of `NUMERIC`/`DECIMAL` mapping**: The old implementation mapped `NUMERIC`/`DECIMAL` declared types to `float64`. Now `sqlite3_column_type()` decides, so a `NUMERIC` column holding an integer value may return `int64` instead.
- **Loss of `CLOB`/`CHAR(N)` mapping**: These declared types were previously recognized as `TEXT` but are now entirely up to `sqlite3_column_type()`.

Also, could you add tests for `BOOLEAN` type (to verify `type_bool` is returned), expressions (e.g. `price * 2`), and `NUMERIC`/`DECIMAL` declared types?

Some of these may be acceptable trade-offs, but I'd like to hear your thoughts.

PR #1344, opened 2025-07-26
Drop userauth implementation

### LINKED ISSUE #1341: sqlite3_user_authenticate was removed
@gjrtimmer 

ext/userauth seems to be removed in https://www.sqlite.org/src/info/59a7eb297cd6aa05

So we have to drop user_auth feature from go-sqlite3. Do you have any idea to avoid this?

mattn:
Closes #1341


sqlite3.go
```diff
@@ -1568,9 +1568,11 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
...
 
-	// Preform Authentication
-	if err := conn.Authenticate(authUser, authPass); err != nil {
-		return nil, err
+	if conn.AuthEnabled() {
+		// Preform Authentication
+		if err := conn.Authenticate(authUser, authPass); err != nil {
+			return nil, err
+		}
 	}
```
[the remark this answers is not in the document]
mattn: No, we need to remain this code for users who uses Authenticate method.

sqlite3_opt_userauth_test.go
```diff
@@ -85,560 +63,14 @@ func init() {
...
 	}
-	defer db2.Close()
-
-	// Delete user while logged in as normal user
-	// through SQL
-	rv, err = deleteUser(db2, "user")
-	if err != nil {
+	if !errors.Is(err, ErrUnauthorized) {
 		t.Fatal(err)
 	}
```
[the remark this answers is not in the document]
mattn: This is a test for connect should return ErrUnauthorized always.

sqlite3_opt_userauth.go
```diff
@@ -131,20 +70,7 @@ func (c *SQLiteConn) authenticate(username, password string) int {
...
-	rv := c.authUserAdd(username, password, isAdmin)
-	switch rv {
-	case C.SQLITE_ERROR, C.SQLITE_AUTH:
-		return ErrAdminRequired
-	case C.SQLITE_OK:
-		return nil
-	default:
-		return c.lastError()
-	}
+	return ErrUnauthorized
```
rittneje, line 73: I think we should return something like `errors.New("sqlite3: the sqlite_userauth tag is no longer supported as the userauth extension is no longer supported by the SQLite authors, see <GitHub issue link>")`.

That way, users that have been relying on this extension will get a very clear message as to what the problem is, as opposed to a mysterious "not authorized" error.

sqlite3.go
```diff
@@ -1568,9 +1568,11 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 		}
 	}
 
-	// Preform Authentication
-	if err := conn.Authenticate(authUser, authPass); err != nil {
-		return nil, err
+	if conn.AuthEnabled() {
```
rittneje, line 1571: I don't think we can add this check. Now `AuthEnabled()` always returns false even if you use the build tag. Therefore, anyone who had been relying on it just skips auth instead of getting the error.

rittneje: Yes. If the user doesn't need this extension, then the fix is simple - stop passing the build tag they don't need.

mattn:
Thank you all.

PR #1350, opened 2025-09-01
Support PRAGMA key

oporat:
Supports PRAGMA key while open db.
That's a must for supporting any sqlite encryption

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1510,7 +1510,7 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 	
 	// _pragma_key
     if encryptionKey != "" {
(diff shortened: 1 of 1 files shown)


sqlite3.go
```diff
@@ -1492,6 +1507,25 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 		C.sqlite3_close_v2(db)
 		return nil, err
 	}
+	
+	// _pragma_key
+    if encryptionKey != "" {
+        query := fmt.Sprintf("PRAGMA key = \"%s\";", pragmaKey)
```
kberov: You can use simply `%q` for readability instead of `\"%s\"`.
See https://pkg.go.dev/fmt#hdr-Printing

kberov (commented):
Great. Thanks!

PR #1352, opened 2025-09-11
Correct handling of SQLITE_TEXT during Change Data Capture

otoolep:


mattn:
>I assume you meant the Copilot review is no good?

Yes, it is.

otoolep:
"Yes it is good and you need to change your code", or "yes, it is no good, the code right now looks good".

PR #1353, opened 2025-09-24
Add function to register wal hook

gfanto:
I have the use case of adding a custom wal hook. 
This PR adds support for registering a [WAL hook](https://www.sqlite.org/c3ref/wal_hook.html) on a connection via a new method:
```go
func (c *SQLiteConn) RegisterWalHook(callback func(dbName string, nPages int) int)
```


gfanto:
> There's a typo in callback.go: `hadlePtr` should be `handlePtr`.

Thanks for the review, I updated the branch and fixed the typo, sorry for the mistake and the late response

Issue #400, opened 2017-03-25
Feature: BEGIN immediate/exclusive per transaction


zombiezen:
Right now, the `_txlock` connection parameter determines which `BEGIN` statement is used to start a transaction.  I'd like to be able to pick this at a per-transaction level.  I don't think there's a way of exposing this functionality to `database/sql`, but it would be good to add an option there eventually.  I'm imagining adding:

```go
func (c *SQLiteConn) BeginWithType(t TransactionType) (driver.Tx, error)

type TransactionType int

const (
  Deferred TransactionType = iota
  Immediate
  Exclusive
)
```

Relevant SO post: https://stackoverflow.com/questions/42657639/begin-immediate-in-sqlx-or-database-sql-golang

mattn:
database/sql has IsolationLevel below.

```go
const (
	LevelDefault IsolationLevel = iota
	LevelReadUncommitted
	LevelReadCommitted
	LevelWriteCommitted
	LevelRepeatableRead
	LevelSnapshot
	LevelSerializable
	LevelLinearizable
)
```
It should map those values into the SQLite's transaction types.

zombiezen:
I don't think it works in this case.  [SQLite's docs][1] assert:

> Except in the case of shared cache database connections with PRAGMA read_uncommitted turned on, all transactions in SQLite show "serializable" isolation.

[1]: https://www.sqlite.org/isolation.html

mattn:
go-sqlite3 is driver for database/sql. So you can't see the BeginWithType from database/sql.

zombiezen:
That's actually something I'm curious about: why does this package export the types used to implement the driver?  Can they be used directly if an application wants to have SQLite3-specific code?

mattn:
I don't make sure but It's possible. However, interfaces are related on database/sql or database/sq/driver.
For example. SQLiteConn#Exec should be passed arguments []driver.Value. So programmer should consider what the value should be passed, what value will be returned.

gaillard:
Well there is https://golang.org/pkg/database/sql/#TxOptions now. But as mentioned its not "isolation level" that is requested but either BEGIN IMMEDIATE/DEFFERED etc. All sqlite transactions are serializable level as far as I understand.
Best I can see to make this possible is for applications have a few different initialized sql drivers at the pkg level that have their own connection hooks and maybe have a field on *sqlite3.SQLiteConn to set?
I would say just change _txlock in the connection hook, but it has to already be given to the sql.Open.

[...]

Issue #685, opened 2019-01-20
BeginTx does not verify TxOptions


rittneje:
The documentation for `ConnBeginTx.BeginTx` says the following:
```go
// This must check opts.Isolation to determine if there is a set
// isolation level. If the driver does not support a non-default
// level and one is set or if there is a non-default isolation level
// that is not supported, an error must be returned.
//
// This must also check opts.ReadOnly to determine if the read-only
// value is true to either set the read-only transaction property if supported
// or return an error if it is not supported.
```

However, this library simply throws away the provided `TxOptions` without verifying them ([link](https://github.com/mattn/go-sqlite3/blob/3fa1c550ffa69b74dc4bfd5109b4e218f32c87cf/sqlite3_go18.go#L50)). This violates the contract. This library must verify that `opts.Isolation` is `sql.LevelDefault` and that `ReadOnly` is `false`.

nevivurn:
+1 to this. Is there some reasoning behind the current behavior, like breaking code that (wrongly) assumes that go-sqlite3 supports these kinds of transactions?

brandond:
Just ran into this myself.

PR #1372, opened 2026-01-24
Add support for the ReadOnly option in BeginTx

jmanero:
This change is similar to the modernc implementation, which ignores the transaction mode defined in the `_txlock` DSN parameter when the `ReadOnly` attribute in `TxOptions` values passed to `BeginTx` is `true`

While the resulting transactions are not actually read-only, this change achieves the result desired by #400 within the confines of the `database/sql` interface.

In `WAL` mode, this allows us to use DSN options like

```
_journal_mode=WAL&_busy_timeout=1000&_txlock=immediate
```

to enable blocking on concurrent calls to `BeginTx(ctx, &sql.TxOptions{})` instead of polling for `SQLITE_BUSY` error codes, while making non-blocking read-only(ish) calls to `BeginTx(ctx, &sql.TxOptions{ReadOnly: true})`.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -962,11 +962,16 @@ func (c *SQLiteConn) query(ctx context.Context, query string, args []driver.Name
 
 // Begin transaction.
 func (c *SQLiteConn) Begin() (driver.Tx, error) {
-	return c.begin(context.Background())
+	return c.begin(context.Background(), false)
 }
 
-func (c *SQLiteConn) begin(ctx context.Context) (driver.Tx, error) {
-	if _, err := c.exec(ctx, c.txlock, nil); err != nil {
+func (c *SQLiteConn) begin(ctx context.Context, ro bool) (driver.Tx, error) {
+	txlock := c.txlock
+	if ro {
+		txlock = "BEGIN"
+	}
+
+	if _, err := c.exec(ctx, txlock, nil); err != nil {
 		return nil, err
 	}
 	return &SQLiteTx{c}, nil
--- a/sqlite3_go18.go
+++ b/sqlite3_go18.go
@@ -40,7 +40,7 @@ func (c *SQLiteConn) PrepareContext(ctx context.Context, query string) (driver.S
 
 // BeginTx implement ConnBeginTx.
 func (c *SQLiteConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
-	return c.begin(ctx)
+	return c.begin(ctx, opts.ReadOnly)
 }
 
 // QueryContext implement QueryerContext.
--- a/sqlite3_go18_test.go
+++ b/sqlite3_go18_test.go
@@ -11,6 +11,7 @@ package sqlite3
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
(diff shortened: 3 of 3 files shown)


MKuranowski:
Any reason why this couldn't be merged? The fact that someone prefers a more convoluted approach which involves going from a simple *sql.DB to something-more-complicated-that-has-multiple-slightly-different-handles shouldn't be any less valid than someone preferring to choose between BEGIN IMMEDIATE and BEGIN DEFERRED transactions.

This doesn't break anything (especially the pooling approach), is very simple and easily explainable for programmers who actually want to use the different transaction modes.

rittneje:
@MKuranowski As previously mentioned, this solution does not actually make the transaction read-only. I really don't think repurposing the flag to mean something else is a good idea.

MKuranowski:
@rittneje I am fully aware of that. But I also don't really see any definition in the sql package that the TxOptions.ReadOnly flag should make data mutations fail. It could as well be treated as a *hint* to the driver that nothing is going to be mutated.

SQLite is already quirky in that it ignores things that would be strict errors in other databases (think `CREATE TABLE numbers (i INTEGER); INSERT INTO numbers VALUES ('foo');`), so this wouldn't be that big of a stretch.

rittneje:
> But I also don't really see any definition in the sql package that the TxOptions.ReadOnly flag should make data mutations fail.

It's literally in the name. The driver package is also quite clear on the matter:

```go
	// This must also check opts.ReadOnly to determine if the read-only
	// value is true to either set the read-only transaction property if supported
	// or return an error if it is not supported.
```

> SQLite is already quirky in that it ignores things that would be strict errors in other databases.

This exact situation (which the SQLite authors tout as a "feature") has been an eternal source of bugs for developers. We should not add another one.

MKuranowski:
I'm still not sold on that. That quote also doesn't imply that an error should be raised when a read-only transaction is upgraded from read-only to read-write. SQLite supports read-only transaction **property** - it's just an optimization for concurrent reads, precisely through `BEGIN DEFERRED`. Side note: under your strict interpretation the driver should raise an error, but it does not.

If you don't like them - don't use them, but don't prevent others from using a genuinely useful feature of SQLite.

rittneje:
> That quote also doesn't imply that an error should be raised when a read-only transaction is upgraded from read-only to read-write.

A deferred transaction is NOT read-only. And SQLite returns an error because it is unable to guarantee the ACID properties of a transaction if the database is modified between the first read and the first write. This has nothing to do with the `ReadOnly` flag.

> SQLite supports read-only transaction property - it's just an optimization for concurrent reads, precisely through BEGIN DEFERRED.

Again, deferred is not the same as read-only. That's what the `query_only` pragma is for.

> Side note: under your strict interpretation the driver should raise an error, but it does not.

That's #685. And it's not a "strict" interpretation - there is no ambiguity in what database/sql/driver says to do.

> If you don't like them - don't use them, but don't prevent others from using a genuinely useful feature of SQLite.

You are arguing to misuse a flag from the standard library and redefine its behavior in a way that violates both the explicit API contract and the principle of least astonishment. If you really want that, you are always free to make a fork.

As far as this PR goes, without enforcing what read-only is supposed to mean, I think it would be a bad idea to take this change. But as I've previously stated, it's reasonable to have it switch to a deferred transaction in _addition_ to enforcing read-only.

Regardless, this conversation seems to be going in circles, so I probably won't be replying any further.

PR #1378, opened 2026-03-18
Add colsUsed to vtable BestIndex callback

### LINKED ISSUE #1323: vtable BestIndex callback missing colUsed input
`struct sqlite3_index_info` includes a mask of all column indexes used: https://github.com/mattn/go-sqlite3/blob/7658c06970ecf5588d8cd930ed1f2de7223f1010/sqlite3-binding.h#L7487

This should be included as a parameter to the BestIndex virtual table callback.

theimpostor:
Split out of #1336 per review feedback.

## Summary
- pass `sqlite3_index_info.colUsed` through to the Go `VTab.BestIndex` callback
- update the examples and tests for the new callback signature
- add test coverage for the propagated `colsUsed` mask

Fixes #1323.

### DIFF
--- a/_example/vtable/vtable.go
+++ b/_example/vtable/vtable.go
@@ -62,7 +62,7 @@ func (v *ghRepoTable) Open() (sqlite3.VTabCursor, error) {
 	return &ghRepoCursor{0, repos}, nil
 }
 
-func (v *ghRepoTable) BestIndex(csts []sqlite3.InfoConstraint, ob []sqlite3.InfoOrderBy) (*sqlite3.IndexResult, error) {
+func (v *ghRepoTable) BestIndex(csts []sqlite3.InfoConstraint, ob []sqlite3.InfoOrderBy, colsUsed uint64) (*sqlite3.IndexResult, error) {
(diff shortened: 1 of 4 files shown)


mattn:
Thanks for splitting this out. The feature itself is welcome — `colUsed` is a useful planner hint and #1323 is a real gap.

My concern is that changing the `VTab.BestIndex` signature is a breaking API change. Every external implementer of `VTab` will fail to compile after upgrading, which we want to avoid in v1.

This repo already uses optional interfaces for extensions (see `VTabUpdater`, `EponymousOnlyModule`). Could you take the same approach here? Something like:

```go
type VTabBestIndexV2 interface {
    BestIndexV2([]InfoConstraint, []InfoOrderBy, uint64) (*IndexResult, error)
}
```

In `goVBestIndex`, type-assert and call `BestIndexV2` if implemented, otherwise fall back to the existing `BestIndex`. That way existing users keep compiling and only those who need `colUsed` opt in.

Minor: SQLite names the field `colUsed` (singular) — matching that naming would make the docs easier to follow.

PR #1387, opened 2026-04-07
[codex] add opt-in statement cache

mattn:
This change adds an opt-in connection-local statement cache to `go-sqlite3`, enabled via the DSN parameter `_stmt_cache_size=N`.

The motivation is that, in this benchmark, the main gap in `mattn/go-sqlite3` was not cgo alone but repeated `prepare/finalize` on hot `QueryRow` paths. Reusing prepared statements at the driver level closes most of that gap without requiring application code to switch to explicit `db.Prepare(...)`.

Implementation details:
- add `_stmt_cache_size` DSN parsing
- keep a small per-connection cache of prepared statements
- reuse cached statements from internal `Query` / `Exec` paths
- reset and return statements to the cache on close
- finalize cached statements when the connection closes
- leave default behavior unchanged unless `_stmt_cache_size` is set

Benchmark workload:
- read-only point lookup
- query: `SELECT id, name, hp, attack FROM monster WHERE id = ?`
- 1000-row SQLite database
- parallel cases:
  - `P1`: about 16 goroutines
  - `P10`: about 160 goroutines
  - `P100`: about 1600 goroutines

Benchmark commands used:

For `mattn/go-sqlite3`:
```sh
go test -run '^$' -bench '^BenchmarkFair(Cached|Prepared|)_C(1|4)_' -benchtime=200ms -count=1
```

For `modernc`:
```sh
go test -run '^$' -bench '^BenchmarkFair(Prepared)?_' -benchtime=200ms -count=1
```

Benchmark results on `2026-04-07` (`Ryzen 7 7735HS`, `WSL2`):

| Condition | mattn before | mattn with stmt cache | modernc |
|---|---:|---:|---:|
| Seq | 10.1 us | 7.7 us | 15.0 us |
| P1 | 42.2 us | 17.7 us | 14.8 us |
| P10 | 42.5 us | 15.7 us | 18.3 us |
| P100 | 45.4 us | 17.7 us | 212.9 us |

In this workload, the cached version improves `mattn` substantially:
- Seq: about 1.3x faster
- P1: about 2.4x faster
- P10: about 2.7x faster
- P100: about 2.6x faster

With the cache enabled, `mattn` is fastest in 3 of 4 measured cases here (`Seq`, `P10`, `P100`), while `modernc` remains slightly faster in `P1`.

This is intentionally opt-in because statement caching changes connection behavior and should be evaluated conservatively in real applications.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -948,7 +948,7 @@ func (c *SQLiteConn) exec(ctx context.Context, query string, args []driver.Named
 
 	start := 0
 	for {
-		s, err := c.prepareWithCache(ctx, query, true)
+		s, err := c.prepareWithCache(ctx, query)
 		if err != nil {
 			return nil, err
 		}
@@ -1013,7 +1013,7 @@ func (c *SQLiteConn) Query(query string, args []driver.Value) (driver.Rows, erro
 func (c *SQLiteConn) query(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
 	start := 0
 	for {
-		s, err := c.prepareWithCache(ctx, query, true)
+		s, err := c.prepareWithCache(ctx, query)
 		if err != nil {
 			return nil, err
 		}
@@ -1969,16 +1969,6 @@ func (c *SQLiteConn) Prepare(query string) (driver.Stmt, error) {
 }
 
 func (c *SQLiteConn) prepare(ctx context.Context, query string) (driver.Stmt, error) {
-	return c.prepareWithCache(ctx, query, false)
-}
-
-func (c *SQLiteConn) prepareWithCache(ctx context.Context, query string, useCache bool) (driver.Stmt, error) {
-	if useCache {
-		if stmt := c.takeCachedStmt(query); stmt != nil {
-			return stmt, nil
-		}
-	}
-
 	pquery := C.CString(query)
 	defer C.free(unsafe.Pointer(pquery))
 	var s *C.sqlite3_stmt
@@ -1992,10 +1982,22 @@ func (c *SQLiteConn) prepareWithCache(ctx context.Context, query string, useCach
 		t = strings.TrimSpace(C.GoString(tail))
 	}
 	ss := &SQLiteStmt{c: c, s: s, t: t}
-	if useCache && t == "" {
+	runtime.SetFinalizer(ss, (*SQLiteStmt).Close)
+	return ss, nil
+}
+
+func (c *SQLiteConn) prepareWithCache(ctx context.Context, query string) (driver.Stmt, error) {
+	if stmt := c.takeCachedStmt(query); stmt != nil {
+		return stmt, nil
+	}
+	stmt, err := c.prepare(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	ss := stmt.(*SQLiteStmt)
+	if ss.t == "" {
 		ss.cacheKey = query
 	}
-	runtime.SetFinalizer(ss, (*SQLiteStmt).Close)
 	return ss, nil
 }
 
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -1884,9 +1884,7 @@ func (c *SQLiteConn) Close() error {
 		return nil
 	}
 	runtime.SetFinalizer(c, nil)
-	if err := c.closeCachedStmtsLocked(); err != nil {
-		return err
-	}
+	c.closeCachedStmtsLocked()
 	rv := C.sqlite3_close_v2(c.db)
 	if rv != C.SQLITE_OK {
 		return lastError(c.db)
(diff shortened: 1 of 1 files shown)


sqlite3.go
```diff
@@ -1883,12 +1905,85 @@ func (c *SQLiteConn) dbConnOpen() bool {
...
 
+func (c *SQLiteConn) takeCachedStmt(query string) *SQLiteStmt {
+	if c == nil || query == "" {
+		return nil
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.db == nil || c.stmtCacheSize <= 0 {
```
rittneje: It seems like `c.stmtCacheSize` is immutable. If so, then we can can check it before the mutex, which avoids the lock overhead for anyone not using this feature.

sqlite3.go
```diff
@@ -2014,11 +2112,26 @@ func (s *SQLiteStmt) Close() error {
...
+	}
 	if !conn.dbConnOpen() {
 		return errors.New("sqlite statement with already closed database connection")
 	}
+	if s.cacheKey != "" {
+		rv := C._sqlite3_reset_clear(stmt)
+		if rv != C.SQLITE_ROW && rv != C.SQLITE_OK && rv != C.SQLITE_DONE {
+			s.s = nil
+			s.c = nil
+			return conn.lastError()
```
rittneje: If we get here, don't we still need to call `sqlite3_finalize`?

sqlite3.go
```diff
@@ -1883,12 +1905,85 @@ func (c *SQLiteConn) dbConnOpen() bool {
...
+
+func (c *SQLiteConn) closeCachedStmtsLocked() error {
+	for key, stmts := range c.stmtCache {
+		for _, s := range stmts {
+			if s == nil || s.s == nil {
+				continue
+			}
+			runtime.SetFinalizer(s, nil)
+			if rv := C.sqlite3_finalize(s.s); rv != C.SQLITE_OK {
+				return lastError(c.db)
```
rittneje: > If the most recent evaluation of the statement encountered no errors or if the statement has never been evaluated, then sqlite3_finalize() returns SQLITE_OK. If the most recent evaluation of statement S failed, then sqlite3_finalize(S) returns the appropriate [error code](https://sqlite.org/rescode.html) or [extended error code](https://sqlite.org/rescode.html#extrc).

I'm not sure we should bail out just because one statement failed to finalize. Especially because it goes on to say "it is a grievous error for the application to try to use a prepared statement after it has been finalized" which means leaving it in the map would be a problem.

sqlite3.go
```diff
@@ -1883,12 +1905,85 @@ func (c *SQLiteConn) dbConnOpen() bool {
...
 // Prepare the query string. Return a new statement.
 func (c *SQLiteConn) Prepare(query string) (driver.Stmt, error) {
 	return c.prepare(context.Background(), query)
 }
 
 func (c *SQLiteConn) prepare(ctx context.Context, query string) (driver.Stmt, error) {
+	return c.prepareWithCache(ctx, query, false)
+}
+
+func (c *SQLiteConn) prepareWithCache(ctx context.Context, query string, useCache bool) (driver.Stmt, error) {
```
rittneje: Instead of the extra boolean, I think it would be simpler to just have `prepareWithCache` call `prepare`.

```go
func (c *SQLiteConn) prepareWithCache(ctx context.Context, query string) (driver.Stmt, error) {
    if stmt := c.takeCachedStmt(query); stmt != nil {
        return stmt, nil
	}

    stmt, err := c.prepare(ctx, query)
    if err != nil {
        return nil, err
    }

    stmt.cacheKey = query
    return stmt, nil
}
```

sqlite3.go
```diff
@@ -1883,12 +1905,85 @@ func (c *SQLiteConn) dbConnOpen() bool {
...
+
+func (c *SQLiteConn) putCachedStmt(s *SQLiteStmt) bool {
+	if c == nil || s == nil || s.s == nil || s.cacheKey == "" {
+		return false
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.db == nil || c.stmtCacheSize <= 0 || c.stmtCacheCount >= c.stmtCacheSize {
```
rittneje: I don't think we need to check `c.stmtCacheSize <= 0` here. Because in that case, it must also hold that `c.stmtCacheCount >= c.stmtCacheSize`.

sqlite3.go
```diff
@@ -2014,11 +2112,26 @@ func (s *SQLiteStmt) Close() error {
...
-	s.c = nil
+	if stmt == nil {
+		s.c = nil
+		return nil
+	}
 	if !conn.dbConnOpen() {
 		return errors.New("sqlite statement with already closed database connection")
 	}
+	if s.cacheKey != "" {
+		rv := C._sqlite3_reset_clear(stmt)
```
rittneje: Should this happen within `putCachedStmt` instead? That way, we skip it if the cache is already full, and also we better guarantee a statement cannot end up in the cache without being reset and cleared?

sqlite3.go
```diff
@@ -1520,6 +1525,17 @@ func (d *SQLiteDriver) Open(dsn string) (driver.Conn, error) {
 			cacheSize = &iv
 		}
 
+		if val := params.Get("_stmt_cache_size"); val != "" {
```
rittneje, line 1531: One concern I have around this that may not be entirely clear to consumers is this is per connection, and `sql.DB` is a connection pool. So if you create a lot of connections - perhaps unintentionally - each has an independent cache.

mattn:
@rittneje Thank you

rittneje:
@mattn Reflecting on this more, it seems like this approach doesn't quite account for evicting stuff from the cache. For example, let's say the cache size is 5. My first 5 queries might be one-time only, and they will all be cached forever. And then the 6th query - which maybe gets run frequently - never gets cached.

mattn:
@rittneje Right. I'll implement LRU cache.

Issue #1390, opened 2026-04-29
Segmentation violation when querying with just a comment


otoolep:
Looks like a simple test coverage issue. Simple query with a comment only, get a panic.

This was fine in earlier versions, and was exposed by rqlite unit testing after upgrading. This test started failing: https://github.com/rqlite/rqlite/blob/v10.0.0/db/db_common_test.go#L109

```bash
~/repos/otoolep-go-sqlite3/src/github.com/otoolep/go-sqlite3 (query-comment-panic)$ git diff
diff --git a/sqlite3_test.go b/sqlite3_test.go
index 852aa10..efc594b 100644
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -2063,6 +2063,11 @@ func TestNamedParamClearBindings(t *testing.T) {
        if z.Valid {
                t.Errorf("Expected z to be NULL, got %d", z.Int64)
        }
+
+       _ = db.QueryRow("---- comment")
 }
 
 var customFunctionOnce sync.Once
~/repos/otoolep-go-sqlite3/src/github.com/otoolep/go-sqlite3 (query-comment-panic)$ go test -run TestNamedParamClearBindings
SIGSEGV: segmentation violation
PC=0x5c7831 m=0 sigcode=1 addr=0x0
signal arrived during cgo execution

goroutine 7 gp=0xc82dc3383c0 m=0 mp=0x8e72e0 [syscall]:
runtime.cgocall(0x5bbb60, 0xc82dc35f5c0)
	/home/philip/.gvm/gos/go1.26.1/src/runtime/cgocall.go:167 +0x4b fp=0xc82dc35f598 sp=0xc82dc35f560 pc=0x48d10b
github.com/mattn/go-sqlite3._Cfunc__sqlite3_reset_clear(0x0)
	_cgo_gotypes.go:401 +0x45 fp=0xc82dc35f5c0 sp=0xc82dc35f598 pc=0x592665
github.com/mattn/go-sqlite3.(*SQLiteStmt).bind.func1(...)
	/home/philip/repos/otoolep-go-sqlite3/src/github.com/otoolep/go-sqlite3/sqlite3.go:2242
github.com/mattn/go-sqlite3.(*SQLiteStmt).bind(0xc82dc3522d0, {0x0, 0x0, 0xc8200000000?})
[...]

PR #1391, opened 2026-04-29
Don't panic when querying with only comments

otoolep:
Fix for https://github.com/mattn/go-sqlite3/issues/1390

Only affects the query path, because the exec path is protected by this: https://github.com/mattn/go-sqlite3/blob/v1.14.42/sqlite3.go#L160

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -2239,6 +2239,9 @@ func stmtArgs(args []driver.NamedValue, start, na int) []driver.NamedValue {
 }
 
 func (s *SQLiteStmt) bind(args []driver.NamedValue) error {
+	if s.s == nil {
+		return nil
+	}
 	rv := C._sqlite3_reset_clear(s.s)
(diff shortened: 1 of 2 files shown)


mattn:
Thanks for the patch!

Unfortunately guarding only `bind()` is not enough for the `QueryRow`/`Query` path. After `bind()` returns, `(*SQLiteStmt).query()` still constructs a `SQLiteRows` whose underlying `s.s` is NULL. When the caller iterates, `(*SQLiteRows).nextSyncLocked()` calls `sqlite3_step(NULL)` (returns `SQLITE_MISUSE`), which is then treated as "got a row" and the subsequent `Scan` produces a confusing error instead of `sql.ErrNoRows`. The `Exec` path you tested already had its own NULL guard via `execNoArgs`, so it never exercised this.

I'll close this in favor of #1392, which detects the NULL statement in `(*SQLiteConn).query()` (mirroring `exec()`) and makes `SQLiteRows` nil-safe so the empty-rows return value behaves correctly.

Thanks again!

otoolep:
All good, main thing is we fix it -- which you did.

PR #1396, opened 2026-05-10
fix: Exec returns wrong arg count in "not enough args" error

### LINKED ISSUE #1395: Exec returns wrong arg count in "not enough args" error
`db.Exec` and `db.Query` share the same multi-statement loop and trip the same "not enough args" check, but the error messages disagree on the second number.

## Repro

```go
_, errExec := db.Exec("SELECT ?; SELECT ?", "hello")
_, errQuery := db.Query("SELECT ?; SELECT ?", "hello")
fmt.Println(errExec)
fmt.Println(errQuery)
```

Output:

```
not enough args to execute query: want 1 got 1
not enough args to execute query: want 1 got 0
```

The two paths use the same check `len(args)-start < na` but format different numbers in the error.

semihbkgr:
Aligns the `Exec` "not enough args" message with `Query`'s. Adds a regression test covering both paths.

Fixes #1395

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected an error message for multi-statement queries so it reports the number of arguments remaining for the current statement.
  * Improved accuracy for both execution and query operations when too few arguments are provided.

* **Tests**
  * Added coverage to verify the corrected error message in relevant query scenarios.

### DIFF
--- a/sqlite3_test.go
+++ b/sqlite3_test.go
@@ -2065,6 +2065,36 @@ func TestNamedParamClearBindings(t *testing.T) {
 	}
 }
 
+func TestNotEnoughArgsErrorMessage(t *testing.T) {
+	db, err := sql.Open("sqlite3", ":memory:")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	const want = "not enough args to execute query: want 1 got 0"
+
+	t.Run("exec", func(t *testing.T) {
+		_, err := db.Exec("SELECT ?; SELECT ?", "hello")
+		if err == nil {
+			t.Fatal("expected error, got nil")
+		}
+		if err.Error() != want {
+			t.Errorf("got %q, want %q", err.Error(), want)
+		}
+	})
+
+	t.Run("query", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


sqlite3_test.go
```diff
@@ -2065,6 +2065,36 @@ func TestNamedParamClearBindings(t *testing.T) {
...
 }
 
+func TestNotEnoughArgsErrorMessage(t *testing.T) {
+	db, err := sql.Open("sqlite3", ":memory:")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	const want = "not enough args to execute query: want 1 got 0"
```
otoolep, line 2220: Unless there are other tests like this in the codebase, this test is brittle and doesn't add much value. The fix is a good idea, but this test is more trouble than it's worth. IMHO.

The right way to test errors is to create type errors and ensure that the right type comes back -- but that is much bigger change and I'm not proposing it.

mattn:
Thank you for reporting this and fixing it yourself, and for covering both the Exec and Query paths in the test!

PR #1398, opened 2026-05-16
Fix wrong arg count in Exec's "not enough args" error

### LINKED ISSUE #1395: Exec returns wrong arg count in "not enough args" error

c-tonneslan:
Closes #1395

`Exec` and `Query` share the same multi-statement loop and trip the same `len(args)-start < na` check, but the error messages disagree on the second number. With one arg passed to `SELECT ?; SELECT ?`:

```
exec  -> not enough args to execute query: want 1 got 1
query -> not enough args to execute query: want 1 got 0
```

The Query path was already right (args remaining after the first statement consumed one), Exec was printing the total args originally passed in. This just brings Exec in line.

Added a regression test that asserts both paths produce the same message.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -962,7 +962,7 @@ func (c *SQLiteConn) exec(ctx context.Context, query string, args []driver.Named
 			na := s.NumInput()
 			if len(args)-start < na {
 				s.Close()
-				return nil, fmt.Errorf("not enough args to execute query: want %d got %d", na, len(args))
+				return nil, fmt.Errorf("not enough args to execute query: want %d got %d", na, len(args)-start)
 			}
 			stmtArgs := stmtArgs(args, start, na)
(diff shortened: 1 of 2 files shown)


mattn:
Thank you for the fix and the well-commented test! #1396 by the original reporter landed first with an equivalent change, so closing this one — the contribution is appreciated.

PR #1399, opened 2026-05-19
avoid out of bounds write in unlock_notify_wait on 64 bit platforms

calmh:
While `C.sizeof_unit` is 4, we convert to a Go uint and write it to the pointer. The Go uint is eight bytes on 64 bit platforms. Allocate the size we intend to write.


mattn:
Thank you

PR #1400, opened 2026-05-19
modernise reflect.SliceHeader to unsafe.Slice

calmh:
Avoid the unsafe use of reflect.StringHeader, using the same pattern already in place in `bindText`, and modernise deprecated reflect.SliceHeader usage to unsafe.Slice.


mattn:
Thank you

PR #1401, opened 2026-06-02
fix: correct inconsistent error message in Exec vs Query (#1395)

vigneshakaviki:
## Summary
- Fixed inconsistent error message reporting in Exec and Query methods
- Both methods validate argument count but reported different values
- Now both report remaining args count after start offset

## Root Cause
Query reported len(args)-start correctly, but Exec reported full len(args)

## Changes
- sqlite3.go line 965: Changed to report len(args)-start instead of len(args)
- Aligns error messages between Exec and Query validation

## Test Plan
- [x] Error messages now consistent between methods
- [x] Reports correct argument count in errors
- [x] All argument validation tests pass

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -962,7 +962,7 @@ func (c *SQLiteConn) exec(ctx context.Context, query string, args []driver.Named
 			na := s.NumInput()
 			if len(args)-start < na {
 				s.Close()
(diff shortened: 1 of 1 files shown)


mattn:
Thank you for the fix! This was resolved by #1396, which also adds a regression test, so closing this one.

PR #1402, opened 2026-06-03
guard oversized string length in ResultText

dxbjavid:
ResultText passes C.int(len(s)) straight through, but ResultBlob already rejects len > math.MaxInt32 on 64-bit. A custom function returning a string over 2 GiB wraps to a negative length, and SQLite then derives the byte count with strlen over the Go string buffer, reading past it. Mirror the existing ResultBlob check so the oversize case returns SQLITE_TOOBIG.


mattn:
Thank you

dxbjavid:
Thanks for the review and merge! Appreciated

PR #1403, opened 2026-06-05
bind via sqlite3_bind_text64/blob64 to avoid 32-bit length truncation

dxbjavid:
reading the bind path I noticed bindText and the blob case in bindValue narrow len through C.int before calling sqlite3_bind_text/blob. on 64-bit a parameter of 2 GiB or more wraps to a negative int32, and sqlite then derives the length via strlen over the go buffer which has no guaranteed nul terminator, so it reads past the end. a 4 GiB value instead truncates to a small positive count and binds the wrong bytes. switching the wrappers to the 64-bit variants keeps the full length so sqlite enforces SQLITE_LIMIT_LENGTH and returns SQLITE_TOOBIG. this mirrors the oversize guard already in ResultBlob/ResultText.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -69,13 +69,13 @@ _sqlite3_open_v2(const char *filename, sqlite3 **ppDb, int flags, const char *zV
 }
 
 static int
-_sqlite3_bind_text(sqlite3_stmt *stmt, int n, char *p, int np) {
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

dxbjavid:
Thanks for the review and merge! Appreciated

PR #1406, opened 2026-06-09
preserve embedded NUL bytes in custom function text values

dxbjavid:
callbackArgString hands a TEXT argument to a custom function by converting it with C.GoString, which stops at the first NUL byte, so a function declared with a string parameter silently loses everything after an embedded NUL in the value it was given; callbackRetText has the mirror problem and returns its string through sqlite3_result_text with a -1 length, so a returned string is cut off at its first NUL on the way back into the database. A function that filters or validates its input then only sees the bytes before the NUL, which can let the rest slip through unchecked. I read the real byte count with sqlite3_value_bytes on the way in and pass the explicit length to sqlite3_result_text on the way out, with the same oversize guard the blob path already carries since we now hand it a real length.

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -208,8 +208,8 @@ func callbackArgString(v *C.sqlite3_value) (reflect.Value, error) {
 		p := (*C.char)(C.sqlite3_value_blob(v))
 		return reflect.ValueOf(C.GoStringN(p, l)), nil
 	case C.SQLITE_TEXT:
-		l := C.sqlite3_value_bytes(v)
 		c := (*C.char)(unsafe.Pointer(C.sqlite3_value_text(v)))
+		l := C.sqlite3_value_bytes(v)
 		return reflect.ValueOf(C.GoStringN(c, l)), nil
 	default:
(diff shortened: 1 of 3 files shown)


mattn:
LGTM. Correctly preserves embedded NULs on both paths, guard matches the blob path, and the test covers both directions. Thanks!

Nit (non-blocking): the TEXT arg case calls `sqlite3_value_bytes` before `sqlite3_value_text`; SQLite's documented order is text-then-bytes. Consistent with the existing BLOB case and harmless on the UTF-8 path, so fine to leave.

mattn:
Ref for the ordering note: https://www.sqlite.org/c3ref/value_blob.html

dxbjavid:
good catch, went ahead and flipped it to text-then-bytes on the TEXT case so it follows the documented order in that ref. left the existing BLOB case as-is since it predates this and isn't really in scope here.

mattn:
Thank you

dxbjavid:
Glad to contribute. Thanks

PR #1408, opened 2026-06-11
free leaked schema string in GetFilename

dxbjavid:
GetFilename hands the result of C.CString straight to sqlite3_db_filename, which only borrows the pointer, so the allocation is never released and a little C heap is leaked on every call. It is the only place in the package that builds a C string for an sqlite3 call without a matching free, the siblings all defer it. Stored the pointer and freed it the same way so the schema name no longer leaks.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -2069,7 +2069,9 @@ func (c *SQLiteConn) GetFilename(schemaName string) string {
 	if schemaName == "" {
 		schemaName = "main"
 	}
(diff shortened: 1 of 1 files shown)


dxbjavid:
Hi maintainers,

Just a friendly follow-up on this PR. When you have a chance, could you please provide an update on its review status?

Happy to make any additional changes if needed.

Thank you!

mattn:
thank you

dxbjavid:
Thank you for your time. Appreciated

PR #1410, opened 2026-06-16
call value_blob/value_text before value_bytes in remaining readers

dxbjavid:
the same call-order issue fixed in callbackArgString (c3e96cd) still affects callbackArgBytes and the preupdate row reader, where sqlite3_value_bytes is invoked before sqlite3_value_blob/sqlite3_value_text. per the documented safe order the count should be read after the pointer, since an internal type or encoding conversion can otherwise leave the length and the buffer describing different representations and the following C.GoBytes over-reads. reordered both blob and text branches in each to fetch the pointer first, matching callbackArgString and the column reader.


dxbjavid:
any update?

dxbjavid:
Closing this pull request due to inactivity

PR #1414, opened 2026-06-30
avoid mutating caller's pass slice in salted crypt encoders

dxbjavid:
the salted variants of the crypt encoders build their hash input with append(pass, salt...). because append reuses the backing array when there is spare capacity, the salt bytes get written past the end of the caller's pass slice and overwrite whatever else happens to live in that array. for a slice carved out of a larger buffer that means a neighbouring slice the caller still holds is silently changed, which is the sort of thing that only shows up intermittently and is awkward to track down on a password path. i ran into it while reading the encoders and reproduced it with a pass slice that shares its backing array with a second slice. switching to a streaming hash, writing pass and then the salt, leaves the digest identical while never writing through the caller's memory. i made the same change to the sha1, sha256, sha384 and sha512 salted encoders since they all shared the pattern, and added a regression test that fails on the old code.


dxbjavid:
Closing this pull request due to inactivity.

PR #1415, opened 2026-07-03
Implement ColumnCount method in SQLiteStmt

alvarolm:
adds a ColumnCount method to SQLiteStmt, following the same idea as the existing Readonly() wrapper

there is nothing new on the C side. The driver already calls sqlite3_column_count inside query() to size the rows struct, so this just exposes a capability that is already there

motivation:

im running arbitrary SQL. before executing a statement, i need to decide whether to call Query or Exec

this pairs nicely with Readonly(), with both methods available. you can inspect what kind of statement you are holding (it writes? return rows?) without actually running anything

---

this seems to be an easy to review pr, there is no new functionality, follows an existing pattern and is just a few lines. maybe im lost, someone can point me in the right direction ?

@mattn 

@otoolep sorry to ping you but it looks like this may be of interest for rqlite also

## Summary by CodeRabbit

* **New Features**
  * Added a public method to report the number of result columns returned by a prepared SQLite statement.
  * Added a public method to retrieve the remaining SQL text after a prepared statement.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -2448,6 +2448,14 @@ func (s *SQLiteStmt) Readonly() bool {
 	return C.sqlite3_stmt_readonly(s.s) == 1
 }
 
+// ColumnCount returns the number of columns in the result set returned by
+// the prepared statement, or 0 if it returns no data (e.g. plain INSERT).
+//
+// See: https://sqlite.org/c3ref/column_count.html
+func (s *SQLiteStmt) ColumnCount() int {
+	return int(C.sqlite3_column_count(s.s))
+}
+
 // Close the rows.
 func (rc *SQLiteRows) Close() error {
 	rc.closemu.Lock()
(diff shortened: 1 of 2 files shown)


otoolep:
Thanks. You should add a unit test.

otoolep:
Create a table, `select *` on it, confirm this function returns 0. Insert X rows, `select *` on it, check the function. Finally delete a row, `select *` on it, confirm it returns previous value-1.

otoolep:
Oh, and check it returns 0 for `INSERT` statements.

alvarolm:
thanks for answering, I understand sqlite3_column_count works over the prepared statement, it just evaluates if the query will return columns, and how many
It has no way to know if there are going be results, so inserting rows does not have any effect

https://github.com/sqlite/sqlite/blob/version-3.53.4/src/vdbeapi.c#L1270

i added a simple test to demonstrate this

this exactly same call: `C.sqlite3_column_count(s.s)` its already being used internally for SQLiteRows https://github.com/mattn/go-sqlite3/blob/master/sqlite3.go#L2358, so in practice is already validated

alvarolm:
added tail retriever method, same as column count, its already being used internally.

https://sqlite.org/c3ref/prepare.html (pzTail)

im using it to filter queries before executing.

this pr has only added methods for calls already used internally

PR #1419, opened 2026-07-13
Return error from vtable cursor open instead of ignoring it

mattn:
cXOpen ignored the error sentinel from goVOpen and returned SQLITE_OK with a nil cursor handle, so a VTabCursor whose Open() fails caused a panic in later callbacks (xFilter/xEof) instead of returning the SQL error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when initializing virtual table cursors.
  * Prevented operations from continuing when cursor creation fails, avoiding potential runtime errors.


mattn:
@coderabbitai review

PR #1446, opened 2026-08-17
fix: keep vtab handle registered when xDestroy fails

canesin:
## Problem
When a virtual table's `VTab.Destroy()` (`xDestroy`) returns an error, SQLite **retains** `p->pVtab` (it only clears the pointer inside `if (rc == SQLITE_OK)` in `sqlite3VtabCallDestroy`) and later calls `xDisconnect` on the same object during unlock/close.

`goVRelease` deleted the Go-side handle regardless of the callback result, so that later `xDisconnect` hit `lookupHandle` -> nil and panicked with a failed type assertion (`interface conversion: interface {} is nil, not *sqlite3.sqliteVTab`) while closing the connection.

## Reproduction (before)

```go
// module whose Destroy() returns an error
db.Exec("CREATE VIRTUAL TABLE vtab USING boom()")
db.Exec("DROP TABLE vtab")        // returns the Destroy error - fine
db.Close()                        // PANIC
```

## Fix
Release the handle only after a successful `xDestroy`; on any `xDisconnect` (success or error) keep releasing as before, since SQLite always discards the object there.

## Test
`TestVTabDestroyErrorThenClose` (build tag `sqlite_vtable`): errors on DROP, then asserts Close completes without panic - it panics on the pre-fix code and passes with the fix. Full `-tags sqlite_vtable` suite green.

Found during the 0xCarbon fork's SQLCipher 4.17 sync review (external review + empirical verification).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved virtual-table cleanup when destruction reports an error.
  - Prevented potential crashes when closing a database after a failed virtual-table removal.

- **Tests**
  - Added regression coverage for failed virtual-table destruction followed by database closure.

### DIFF
--- a/sqlite3_opt_vtable.go
+++ b/sqlite3_opt_vtable.go
@@ -423,8 +423,15 @@ func goVRelease(pVTab unsafe.Pointer, isDestroy C.int) *C.char {
 	} else {
 		err = vt.vTab.Disconnect()
 	}
-	// The vtab is gone as far as SQLite is concerned regardless of the
-	// callback result, so release the handle either way.
+	if err != nil && isDestroy == 1 {
+		// SQLite retains p->pVtab when xDestroy fails and later calls
+		// xDisconnect on the same object; keep the handle registered so
+		// that call can resolve it (a missing handle panics above).
+		return mPrintf("%s", err.Error())
+	}
+	// On xDisconnect (success or error) SQLite discards the vtab object,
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

mattn:
Thank you for the thorough report and fix! The reproduction against sqlite3VtabCallDestroy semantics was spot on. Released in v1.14.50.

PR #1452, opened 2026-09-05
Flush statement cache when the schema changes

### LINKED ISSUE #1447: stmt cache (_stmt_cache_size) returns expired statements after schema changes (wrong results)
## Defect
With `_stmt_cache_size` enabled, the statement cache returns cached statements that a schema change has expired, producing silently wrong results.

## Reproduction
```go
// open with ?_stmt_cache_size=8
db.Exec("CREATE TABLE t (a TEXT)")
db.Query("SELECT * FROM t").Close()     // cached
db.Exec("ALTER TABLE t ADD COLUMN b TEXT")
rows, _ := db.Query("SELECT * FROM t")  // cache hit
cols, _ := rows.Columns()
// cols == [a]  <- b is silently missing
```

## Cause
`takeCachedStmt` returns the cached `sqlite3_stmt` without checking `sqlite3_stmt_expired()` (or the schema cookie). SQLite re-prepares on the first `sqlite3_step` after the schema change, but `SQLiteRows.nc`/column metadata were captured at `query()` time from the expired statement, so the Go side reports the old columns and silently omits new ones.

Observed at current master (stmt cache landed in #1387 lineage); reproduced empirically by the 0xCarbon fork's sync review.

## Suggested directions
- check `sqlite3_stmt_expired()` on cache reuse and re-prepare when set, or
- capture `nc`/column metadata lazily on first step, or
- invalidate the cache on schema-cookie change.

Happy to send a PR if you prefer a specific direction.

mattn:
Fixes #1447. With _stmt_cache_size enabled, a schema change expired cached statements but the cache still served them; the column count and metadata captured before stepping described the old schema, so e.g. SELECT * silently omitted newly added columns.

Each connection with the cache enabled now holds a pre-prepared probe statement (SELECT 1 FROM sqlite_master LIMIT 0). Stepping it on every cache lookup makes SQLite verify the schema cookie and reload the connection's schema after a change by another connection, and its SQLITE_STMTSTATUS_REPREPARE count reveals that a change happened, in one CGO crossing. On a change the whole cache is flushed (every cached statement is expired anyway). If the runtime library is too old to report re-prepare counts (pre-3.20), the cache is disabled instead of risking stale results.

Covers both same-connection DDL and DDL from a second connection in the regression test. Cache-hit cost goes from ~820ns to ~1016ns; still well below the ~2.1µs uncached path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved prepared-statement cache handling when database schema changes occur.
  - Automatically refreshes cached statements so updated column metadata and query results are used.
  - Improved handling of schema changes involving temporary tables and explicit transactions, including after commit.
  - Prevented statement-cache issues when running with older SQLite library versions by disabling caching when necessary.

### DIFF
--- a/sqlite3.go
+++ b/sqlite3.go
@@ -135,6 +135,28 @@ _sqlite3_reset_clear(sqlite3_stmt* stmt)
   return rv;
 }
 
+// Steps the pre-prepared schema probe statement and returns its
+// cumulative re-prepare count, in a single CGO crossing. The probe
+// references the schema, so stepping it makes SQLite verify the schema
+// cookie and reload the connection's schema after a change by another
+// connection; the re-prepare count then reflects that change.
+// Returns -1 on error or when sqlite3_stmt_status is too old to
+// report re-prepares.
+static sqlite3_int64
+_sqlite3_schema_generation(sqlite3_stmt* stmt)
+{
+#ifdef SQLITE_STMTSTATUS_REPREPARE
+  int rc = sqlite3_step(stmt);
+  sqlite3_reset(stmt);
+  if (rc != SQLITE_ROW && rc != SQLITE_DONE) {
+    return -1;
+  }
+  return sqlite3_stmt_status(stmt, SQLITE_STMTSTATUS_REPREPARE, 0);
+#else
+  return -1;
+#endif
+}
+
 #ifdef SQLITE_ENABLE_UNLOCK_NOTIFY
 extern int _sqlite3_step_blocking(sqlite3_stmt *stmt);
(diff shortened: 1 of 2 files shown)


mattn:
Addressed the review: the cache is now disabled when the runtime library predates SQLITE_STMTSTATUS_REPREPARE (3.20), the probe also references sqlite_temp_master, lookups inside an explicit transaction report a miss instead of stepping the probe (avoiding snapshot widening), and the reuse test skips when the cache is disabled. Added regressions for temp-schema DDL and DDL inside a transaction. Statements referencing ATTACHed schemas remain uncovered; documented at the probe.

mattn:
@coderabbitai review

